PR #10, opened 2016-03-11
Fixed tests 3 & 4, added test for keeping value selected when multiple=false and options change

owenconti:
screen shot 2016-03-10 at 8 45 54 pm

The code in the watch:options function is a bit long, but it's hard to cover both multiple=true and multiple=false use cases in one function.

I also wasn't sure if this functionality was already in place somewhere but just wasn't working. Let me know if that's the case and I can update this


benkelukas:
Hi, will this be merged? I'm having an use case where this causes trouble (e.g. after adding to options, already selected values gets emptied)

benkelukas:
@sagalbot Thanks for letting me know, awesome work by the way, thanks both of you :)

PR #11, opened 2016-03-11
Changed 'selected' in README to 'value'

owenconti:


owenconti:
Did not understand parent component reference.

PR #13, opened 2016-03-11
Added rich text option by outputting HTML instead of plaintext

owenconti:
screen shot 2016-03-10 at 9 48 06 pm


owenconti:
Hmm yeah I read that bit about XSS on the vue website but I can't think of anything either, especially because this is only for display.

I can see issues where certain HTML output could break the select component, but it would probably be the responsibility of the implementing developer to prevent that or sanitize the output provided.

PR #16, opened 2016-03-12
Added support for themes

owenconti:
Previous styles:
screen shot 2016-03-11 at 6 54 48 pm

New Cyan theme:
screen shot 2016-03-11 at 6 18 24 pm

New Default theme:
screen shot 2016-03-11 at 6 16 41 pm


src/components/Select.vue
```diff
@@ -85,26 +82,61 @@
...
+
+  .theme-cyan.dropdown.open .dropdown-toggle,
+  .theme-cyan.dropdown.open .dropdown-menu {
+    border-color: #4CC3D9;
+  }
+
+  .theme-cyan .active a {
     background: rgba(50,50,50,.1);
     color: #333;
   }
```
sagalbot: This chunk of theme styles can go to `App.vue`

owenconti:
Okay no problem, I can work on these. I have a few questions/concerns though:

1. If I move the .theme-cyan styles into App.vue, the demo page will show the component styled with the the cyan theme, but when the user downloads and starts using the component they will have no theme. Kind of like false advertising. Should there be some sort of UI toggle on App.vue to toggle the styles?

2. If the concern is to keep sizes and dependencies to a minimum, should we look at removing the dependency on bootstrap? I see animate.css is also included in App.vue, but I don't know if/where it is being used. 

This is what the page looked like without the bootstrap stylesheet:

screen shot 2016-03-14 at 5 57 28 am

owenconti:
Updated the PR. It is showing 10,526 deletions in the build.js, hopefully everything is alright.

PR #23, opened 2016-03-24
Added option for letting the user create options from search 

nanotronic:


Frozire:
+1

nanotronic:
Will do. There is one thing still missing. The current search input isn't displayed in the dropdown.

PR #93, opened 2016-09-21
Added possibility of changing de message appearing when not finding options

LuanAraldi:
Added possibility of changing de message appearing when not finding any option on autocomplete, being default as "Sorry, no matching options.". Can be changed by passing the prop "no-option-msg" on the component.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -217,7 +217,7 @@
 			</li>
 			<li transition="fade" v-if="!filteredOptions.length" class="divider"></li>
 			<li transition="fade" v-if="!filteredOptions.length" class="text-center">
-				<slot name="no-options">Sorry, no matching options.</slot>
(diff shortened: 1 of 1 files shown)


LuanAraldi:
Sorry, haven't found that on the docs, so decided to do it myself.

phaberest:
Does not seem to work with latest version. Copy pasted your `span` but can't get it to work.

```
<v-select :on-search="getDestinationOptions"
          :options="destination_search_options"
          :debounce="250"
          :on-change="geoDestinationSelected"
          :reset-on-options-change="clear_search"
          :value.sync="destination_selected"
           @keydown.enter.prevent=""
           v-if="(! shipment.receiver.city) && (shipment.receiver.country == 'IT')"
           label="label"
           placeholder="CAP"
>
  <span slot="no-options">This should be the message.</span>
</v-select>
```

PR #96, opened 2016-09-29
Changed implementation of non searchable option.

### LINKED ISSUE #97: Tabbing, keyboard navigation and focusability unavailable when searchable=false
In the current implementation of vue-select when you make the component `searchable=false` it uses a `span` instead of the main `input` field. This causes several issues as the `span` element  is not focusable causing tabbing to be disabled and you also lose the standard keyboard navigation that you normally have in searchable mode.

Another side effect of the current implementation is that styling needs to be modified for the non searchable elements to make them look the same.

mofodojodino:
This now uses the same input field but makes it readonly instead of trying to swap it out for a non focusable element in turn causing the following issues:
* styling is different
* element is not focusable
* keyboard support - navigating options is disabled
* cannot tab to component

Fixes #97


purinda:
👍

mofodojodino:
No worries, glad I could help.

PR #100, opened 2016-10-20
Added support for Vue2.0

akkinenirajesh:
There is still problem with the tests.


akkinenirajesh:
I could get up to
Executed 50 of 50 (17 FAILED) (0.144 secs / 0.301 secs)
TOTAL: 17 FAILED, 33 SUCCESS

The major change for us is that vue 2.0 does not let us modify properties. 
https://rc.vuejs.org/guide/migration.html#Prop-Mutation-deprecated

So we can not have 'value' as property to set the data and use the same to get the databack. Only way to get the data back is to use another data property (something like 'currentValue') 

So we need to decide what should be the name of 'value' property and what should be the name of data property. They both can not be same name.

What names do you suggest?

PR #101, opened 2016-10-20
More changes for Vue2.0

akkinenirajesh:
The way test code is checking for callback is not working. 
But actual callback are working in code.


akkinenirajesh:
8 our 50 tests are failing

PR #113, opened 2016-12-09
fix event handling

tonypee:
It seems that there could not be two event handlers on the same element, for the same key, and vue 2 does not come with any built in filters, so ive used the native filter.
eg:
```
@keyUp="handle"
@keyUp.enter="handleEnter"
```
the enter was always being called

this seems to make the select work as expected (until i find more bugs and fix)


src/components/Select.vue
```diff
@@ -620,7 +637,7 @@
 			 * @return {array}
 			 */
 			filteredOptions() {
-				let options = this.$options.filters.filterBy?this.$options.filters.filterBy(this.currentOptions, this.search):this.currentOptions
+				const options = this.currentOptions.filter(option => ~option.indexOf(this.search))
```
zweifisch: what if the option is an object?

sagalbot: @zweifisch thanks for the QA!

PR #129, opened 2017-02-10
Add ability to add id to the input element

mgmeiner:
Hey,

made this small PR to add the ability to set the id of the input element. I needed this to focus the vue-select component via labels.

Example:

`<label for="selectCountry">Country:</label>`
`<vue-select input-id="selectCountry" />`


boycce:
Maybe we should add this into the docs?

PR #171, opened 2017-03-30
Added features to non multiple selects

fidgetwidget:
- added a single class to the selected-tag for non multiple selects
- added styling to the single selected-tag to differentiate it from the default

image

image

### DIFF
--- a/dev.html
+++ b/dev.html
@@ -31,6 +31,7 @@
 
 <body>
     <div id="app">
+        <v-select placeholder="default" :options="options"></v-select>
         <v-select placeholder="multiple" multiple :options="options"></v-select>
         <v-select placeholder="multiple, taggable" multiple taggable :options="options" no-drop></v-select>
         <v-select placeholder="multiple, taggable, push-tags" multiple push-tags taggable :options="[{label: 'Foo', value: 'foo'}]"></v-select>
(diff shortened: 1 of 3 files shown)


fidgetwidget:
I'll happily make those changes.
Do you think it would be better to add a class of 'searching' (or something similar) when there is content in the input, or just put a v-if/v-show on the selected-tag?

fidgetwidget:
to ensure uses who prefer it's current behaviour, a css focused approach makes more sense imo, so that's what I'll do here. 👍

fidgetwidget:
I've also added a minor addition
- when clearOnSelect is true and the select is a single (so the default state), the search text is cleared on blur.
This feels more in line with what mine and a few others I ran it by expected than the alternative (which is also an option if you turn clearOnSelect off)

fidgetwidget:
Looking at the contribution guide, I'll try and write some coverage for the new functionality today, so that it's a better PR

AxelTheGerman:
Amazing. I was just looking for a non jquery replacement for select2/selectize and I love Vue... first thing that stood out to me at the demo page was the UI for single select just being the same as for multiple.
Thanks for the PR, hope to see it getting released soon 👍

fidgetwidget:
@AxelTheGerman me too! :D
@sagalbot Any eta?

PR #172, opened 2017-03-31
Add option to allow closing a multi-select when a value is selected

koriroys:
Hello!

My boss wants the option to close the dropdown when a single value is selected, even if it is a multi-selection dropdown (Since the most likely use case is to choose one thing, even if it is allowed to choose more than one). Thought I'd send the change upstream to see what you think. I'm not sure this is the best name, or even best way to add this option, so maybe you have some ideas?

Cheers!

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -391,7 +391,7 @@
       },
 
       /**
-       * Close a dropdown when an option is select. Set to false to keep the dropdown
+       * Close a dropdown when an option is chosen. Set to false to keep the dropdown
        * open (useful when combined with multi-select, for example)
        * @type {Boolean}
        */
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -391,6 +391,16 @@
       },
 
       /**
+       * Allows user to choose to close the select dropdown when an option is selected.
+       * set to true when multiple=true to close the dropdown between each selection
+       * @type {Boolean}
+       */
+      closeOnMultiSelect: {
+        type: Boolean,
+        default: false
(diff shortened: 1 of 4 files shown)


src/components/Select.vue
```diff
@@ -391,6 +391,16 @@
       },
 
       /**
+       * Allows user to choose to close the select dropdown when an option is selected.
+       * set to true when multiple=true to close the dropdown between each selection
+       * @type {Boolean}
+       */
+      closeOnMultiSelect: {
```
sagalbot: I think we can simplify this to be `closeOnSelect`.

src/components/Select.vue
```diff
@@ -391,6 +391,16 @@
...
+        type: Boolean,
+        default: false
```
sagalbot: I wonder if this should be `true` by default. I took a look at Select2, and that's the default functionality there.

src/components/Select.vue
```diff
@@ -625,7 +635,12 @@
        * @return {void}
        */
       onAfterSelect(option) {
```
sagalbot: This can probably be refactored to be quite simple:

```js
if (this.closeOnSelect) {
    this.open = !this.open
    this.$refs.search.blur()
}
```

koriroys:
Should this be true by default for multiple selection? That would change the current behavior.

koriroys:
@sagalbot thoughts?

koriroys:
@sagalbot you're welcome! It's been a pleasure. 🎉🎉🎉

PR #174, opened 2017-04-01
Changing text such as 'Loading...' and 'Sorry, no matching options' using properties

ik5:
The following version adds the ability to change the text that currently are fixed.
I think that the placeholder also needs to add there, but that will break current code.


ik5:
that's cool, didn't notice it, thank you :)

PR #185, opened 2017-04-18
Adding support to filter results by value

Jonas-buriti:
This commit implements a new functionality which allows the user searching by the value (besides the label). The boolean property "filterByValue" controls this behaviour.


msyesyan:
@sagalbot this is a good feature, why not we merge this branch?

Jonas-buriti:
@sagalbot it makes sense!
I've made another commit which implements a filterBy property.

Issue #190, opened 2017-04-23
Ability to specify custom filteredOptions function


stereosteve:
Current [filteredOptions](https://github.com/sagalbot/vue-select/blob/aabc43c903a8caf8ff8116304bcceedb4693e33b/src/components/Select.vue#L810-L823) function provides a nice default string search.

I'm interested in doing some pre-processing of user input + custom filter behavior.  What do you think about providing an API option to override default `filterOptions`?

sagalbot:
Absolutely agree. `filteredOptions` should really be abstracted to a prop callback with the current implementation as a default. Additionally, I want to add another prop that allows you to use the default filtering, but specify your own key. Right now it just looks at your labels.

PR #185 suggested adding a `filterByValue` prop as well that would check for `object.value` and filter against that, but that assumes your data has `.value` property, so a more flexible implementation would be to instead specify a `filterKey`.

sagalbot:
If you're interested in working on it, just let me know, and I'll create a feature branch for it that you can submit a PR against.

stereosteve:
Yeah I'd be happy to work on this.

sagalbot:
Awesome! It's something I've wanted for a while but have other stuff on the plate.

I just pushed up a new `improved-filtering` branch. You can submit PRs against that branch. I'll close this issue but feel free to keep the conversation going here.

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [Customizable filtering](https://trello.com/c/KforHZFo/21-customizable-filtering)

PR #191, opened 2017-04-23
filter-key and filter-callback props

stereosteve:
See https://github.com/sagalbot/vue-select/issues/190 for initial discussion.

### DIFF
--- a/dev.html
+++ b/dev.html
@@ -40,7 +40,7 @@
         <v-select placeholder="unsearchable" :options="options" :searchable="false"></v-select>
         <v-select placeholder="search github.." label="full_name" @search="search" :options="ajaxRes"></v-select>
         <v-select placeholder="custom filter key" multiple :options="options" filter-key="value"></v-select>
-        <v-select placeholder="custom filter function" multiple :options="options" :filter-callback="customFilter"></v-select>
+        <v-select placeholder="custom filter function" multiple :options="options" :filter="ignoreSpaceMatcher"></v-select>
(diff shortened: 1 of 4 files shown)


stereosteve:
@sagalbot can you review?

I rebased on master, let me know if that causes any problems.

stereosteve:
Sounds good, no rush to merge.  I'm considering another option:

* If option has `hidden = true` it is not shown
* Emit events when search input changes

Calling code can listen for search change events (or anything else really) and change what options are shown + hidden.  Also will have access to option object so can use extra meta data.

Going to give it a try.

stereosteve:
@sagalbot I tried a few things, but they were too complicated.  I think this PR is good enough.

I renamed `filter-callback` to just `filter`, and swapped the parameters so that the filter value is the first parameter to mimic standard JS filter function.  `option` is passed as 3rd parameter in case you wish to use additional properties of the option for filtering.

Let me know if this looks okay.  I can squash to 1 commit if you prefer.

alan-zhang-:
Hi, can this feature be used for now? I definitely need to stop client side filtering.

Snippo:
I also needed to disable the filter functionality. You can install the pull request with npm like this:
npm i --save-dev sagalbot/vue-select#pull/191/head

sdmackay:
Any news on when this is going to be merged into master?

sdmackay:
@sagalbot Is this PR and #233 doing the same things?

PR #118, opened 2016-12-26
Add option to disable the entire component

evanslify:
Issue #116 

Added functionality to disable vue-select by adding `disabled` attribute to the component. 
Note that the docs is broken for now so that I can't add description to docs.

PR #208, opened 2017-05-09
Allow disable

PauloFerreira25:
Hello,
Sorry if I did stupidly.
It was the best I could do. :(
Please accept and / or improve.
Thanks for listening


ecmel:
Hi there, there is an existing PR for disabling entire component (https://github.com/sagalbot/vue-select/pull/118), is this different?

PauloFerreira25:
Hi,
Sorry, I did not see that there was such a request. My fault.
Looking at the request, I believe it's the same thing.

image

In the image, I can not capture the mouse, but it puts the denied pointer.
Please see what is best to do, use mine or discard it.
Sorry for the inconvenience and thanks for your attention.

TiagoSilvaPereira:
Hi,
Is this feeature in the master? I can't use it in my implementation.
Thanks

Issue #30, opened 2016-04-13
How can we theme the vSelect


nacr-dev:
Is there any support for themes?

And if so how can it be done?

tks

sagalbot:
There aren't any pre-made themes at the moment. We discussed it in PR #16 but decided that it was unlikely the themes we produce are going to line up with a given users design. If I was wrong about that, I'm happy to merge #16. At the moment, your best bet would be write your own styles from an external stylesheet. 

What are you looking for in terms of themes?

sagalbot:
For example, on the demo site, I've just added an ID to the element and add some simple styles to change the colors a bit:

```css
/* Cyan theme */
#v-select .selected-tag {
    color: #147688;
    background-color: #d7f3f9;
    border-color: #91ddec;
}

#v-select.dropdown.open .dropdown-toggle,
#v-select.dropdown.open .dropdown-menu {
    border-color: #4CC3D9;
}

#v-select .active a {
    background: rgba(50,50,50,.1);
    color: #333;
}

#v-select.dropdown .highlight a,
#v-select.dropdown li:hover a {
    background: #4CC3D9;
    color: #fff;
}
```

nacr-dev:
Ok will take a look.

Some how I did not get this working yesterday.

nacr-dev:
ok Face Palme for me :)

tks m:8ball:

cleitonsouza:
Sorry for re opening this issue

I took a look in PR #16, but i got the following conclusion: to add my own style for this component, i need to override the current style. I'm right?

If so, then i have a suggestion: add a new prop, like `theme`, and define a default theme, that have the current style. This will add a classe in the root element `default-theme`, and the style will be writen like: `.default-theme .dropdown-toggle`, etc..

In this way, when we pass `<v-select theme="my-own">`, we can develop the style from scratch

Thanks

PR #217, opened 2017-05-17
Allow to build own theme

cleitonsouza:
This PR create the `theme` prop.

This prop will bind theme `theme-${theme}` class in the root element (`.v-select`), with default value as "default". 
All styles defined in the component is now inherited from `.v-select.theme-default`.

So, this allows to create a own theme.
If you define the theme="foo", you can build your theme from scratch, without overrides, using `.v-select.theme-foo`.

I've mentioned this in the issue #30.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -1,16 +1,16 @@
 <style>
-  .v-select {
+  .v-select.theme-default {
     position: relative;
     font-family: sans-serif;
   }
-  .v-select,
-  .v-select * {
+  .v-select.theme-default,
+  .v-select.theme-default * {
     -webkit-box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


cleitonsouza:
The only css that remains is about transition.
In my opnion, no css must keep.

So, i think we can offer the css of the "default" theme, so people can write from it.

I have this in scss version (i've converted).
What you think about a folder to put the community versions?

gabrielbuzziv:
+1

diego3g:
+1 I also need this.

ecmel:
As long as the default theme is consistent with bootstrap 4;+1

cleitonsouza:
Great point of view :smile:

Issue #106, opened 2016-11-09
In IE, clicking the scrollbar collapses the dropdown


pascallaliberte:
Tested in IE 10, IE 11:

When clicking on the scrollbar of the dropdown menu, the dropdown collapses unexpectedly. See the gif:

![vue-select-scrollbar-ie](https://cloud.githubusercontent.com/assets/104179/20146558/6734ae46-a672-11e6-960b-7186fb90402b.gif)

pascallaliberte:
I'm running into a similar issue elsewhere, so I hope this helps debug it: in IE, clicking a scroll bar will bubble up a click event further up the parent chain. Maybe there's a way to limit the click's reach, but I haven't found one yet.

wa05:
Same problem in Chrome...

Saul3d:
I'm not seeing this problem in chrome but does anyone have a temporary fix for this issue in IE

andremiguelaa:
Any news on this issue?

eckledes:
Is this bug fixed in Release 2.3.0?

adi518:
@eckledes No.

adi518:
Okay, should be fixed now: #373

korobkadima:
same problem on Chrome in modal window

nahuelhds:
I'm on v2.5.1 and this keeps happening. In my case, only the first time I try to scroll with native browser UI after doing an ajax search with brings me lot of results. Once closed, if I open the dropdown again then I can scroll using the native browser UI.

PR #229, opened 2017-06-01
fix bug #106

andremiguelaa:


adi518:
I'm surprised this held so long in under PR. It's annoying bug. Why did you use `setTimeout`?

adi518:
Fixed here: #373

PR #233, opened 2017-06-13
Added option :filterable

Zhomart:
Sometimes we don't want to filter options by label, the options are already filtered on the server side.

If the option `filterable` is set to `false`, vue-select won't filter options and will show all options as is.

Added tests. Here is an example:

```
<v-select ref="select" :filterable="false" :options="['foo','bar','baz']" v-model="value"></v-select>
```

I can add docs if needed.


francois2metz:
Could be closed. Added in https://github.com/sagalbot/vue-select/commit/2bf2661405e6f5cf32b6ba3f2bf62ce64248532c

PR #249, opened 2017-07-05
Prevent submitting inside form

SpyTec:
I discovered that when vue-select is inside a form it does not prevent a submit when hitting enter. Having a guess since it prevents on `keyup` instead of `keydown` - As forms submit when pressing enter, not when releasing.


mrdoinel:
I think this is quite an import blocker when vue-select is used for tagging.
Any updates on this PR ?

PR #250, opened 2017-07-06
Added support for tab index

chiragparekh:
I have added tab index attribute to component so user can set custom tab index.


chiragparekh:
I just closed it will it work?

PR #257, opened 2017-07-12
Emit 'reset' event on empty search

erik-sjoestedt:
I would like a 'reset' event for when the search result is set back to being empty.


erik-sjoestedt:
Always emitting the search change sounds good to me, as long as one can detect "empty" searches aswell

valeriob:
Hi guys, can this pull request be merged ?

erik-sjoestedt:
I changed the PR to always emitting the search event, even on empty strings, as suggested by @sagalbot .

PR #261, opened 2017-07-19
Initially "Start Searching." <= "Sorry, no match*"

sakshigargit:
When we click on input box of vue-select it always say "Sorry, no matching options"... whereas there is no value in the input box to say this statement it should say "Start Searching...". So, from in the tag commit I have changed the text from "Sorry, no matching options" to "Start Searching..." in case there is no value in input box

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -309,8 +309,11 @@
             {{ getOptionLabel(option) }}
           </a>
         </li>
-        <li v-if="!filteredOptions.length" class="no-options">
+        <li v-if="!filteredOptions.length && search" class="no-options">
           <slot name="no-options">Sorry, no matching options.</slot>
+        </li>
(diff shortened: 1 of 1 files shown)


sakshigargit:
@sagalbot Please, can you make this approved. I really need this in my project. I hope this will follow a proper approach everyone required

sakshigargit:
@sagalbot Can you please also look how can I update vue-select.js with the current things I have changed in select.vue file as I am using that

Arteyan:
Could these not instead be added as props?
As someone who utilizes Vue with Laravel it would be helpful for various locale if these could be customized instead of hard coded.

sakshigargit:
Yes, it can be. So, I will update it to props and make it next patch. But for now, it is really required to be merged and released.

PR #266, opened 2017-07-21
Fixed behavior of the "active" class when using an object as value

### LINKED ISSUE #263: Fix "active" class behaviour on non String default values
If we pass a string value to the "value" property, the component will add the "active" class to the corresponding option:

screenshot from 2017-07-19 13-20-40

But when we try to pass an Object, it doesn't:

screenshot from 2017-07-19 13-21-42

emil14:
I want to resolve #263 

The method "isOptionSelected" added a check for the case if user pass an
Object to "value" property. In this case, for comparison it is more
correct to use the property "label".

### DIFF
--- a/dev.html
+++ b/dev.html
@@ -31,14 +31,14 @@
 
 <body>
     <div id="app">
-        <v-select placeholder="default" :options="options"></v-select>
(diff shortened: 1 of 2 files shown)


emil14:
I apologize, forgot to run the tests. I will try to fix PR soon

Issue #39, opened 2016-05-29
[Question ] RTL?


mosid:
Is right to left feature supported?

sagalbot:
Not yet, but I'll get on it asap and let you know.

adi518:
Preliminary RTL support based on default usage:

```
/* rtl */

.v-select .open-indicator {
  left: 10px;
  /* override */
  right: auto;
}

.v-select .selected-tag {
  float: right;
  margin: 4px 3px 0px 1px;
}

.v-select .dropdown-menu {
  text-align: right;
}
```

adi518:
I forked it and added both rtl and `i18n` support with a fallback. Should PR that tomorrow.

sagalbot:
@adi518 PR for that would be very appreciated! This has been on my list for a long time.

adi518:
Sweet. Great to see you're still active on this.

sagalbot:
@adi518 need to be much more active, but doing what I can. I moved jobs a couple months ago and haven't been able to give it the time it deserves. Looks like things will slow down in September and I'm hoping to be back to full steam. I have a lot planned for the component.

adi518:
I PR'd RTL, but left localization out for the time being. I'm a little new to Vue and its best practices.

meteorlxy:
As for localization support, #297 may help.

adi518:
I'll look into that soon. Btw, I managed to fix few other things by extending the component with Vue `extends` composition property, allowing me to skip PRs for the time being.

sagalbot:
Done https://github.com/sagalbot/vue-select/releases/tag/v2.3.0

PR #274, opened 2017-08-02
Add RTL support

adi518:
Adds #39


src/components/Select.vue
```diff
@@ -796,7 +818,8 @@
           searching: this.searching,
           searchable: this.searchable,
           unsearchable: !this.searchable,
-          loading: this.mutableLoading
+          loading: this.mutableLoading,
+          rtl: this.rtl
```
sagalbot: No need for this here. The data properties above it are all modified internally by the component. Since props are immutable, they are proxied to mutable values here.

Since vue-select has no reason to change `rtl`, we don't need to create a mutable data property for it.

adi518: Hmm, I don't recall making it mutable.

sagalbot: You have it defined as both a prop, and a data key.

adi518: Really strange, you'd get an error from Vue. IIRC, I set it as prop and then added it to the `dropdownClasses` computed prop.

sagalbot: Whoah! I've reviewed too many PRs today.. The way github presented the changes totally makes it look like it was in `data`, but yeah, it's in `dropdownClasses()`.

adi518: Yes, it appeared like that for me as well. Kinda confusing, mysterious ways of the Github.

adi518:
I assigned the `dir` attribute. I don't think you can shorthand `rtl`. It'd be nice though. :)

adi518:
Oh. Didn't know that one. Slick. Silly me. Didn't notice the `auto` option of DIR attribute, as documented in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir).

PR #295, opened 2017-08-26
Fix bug if the option label is number.

meteorlxy:
I met this bug when the `typeof` my label is `number`, as there is no `toLowerCase()` method on `NumberObject`.
```
[Vue warn]: Error in getter for watcher "filteredOptions": "TypeError: e[t.label].toLowerCase is not a function"
```


mariojankovic:
Any clues on when this will be merged?

PR #297, opened 2017-08-26
Add loadingText and noMatchText.

meteorlxy:
For localization&customization consideration. Then we can use custom text in our own language by `props`.

Example:
```html
<v-select loading-text="加载中..." no-match-text="没有匹配结果"></v-select>
```

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -298,7 +298,7 @@
       <i v-if="!noDrop" ref="openIndicator" role="presentation" class="open-indicator"></i>
 
       <slot name="spinner">
-        <div class="spinner" v-show="mutableLoading">Loading...</div>
+        <div class="spinner" v-show="mutableLoading">{{ loadingText }}</div>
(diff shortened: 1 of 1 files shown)


meteorlxy:
OK I'll revert them. Maybe the slot should be modified, too. I'll update the pr soon

meteorlxy:
@sagalbot And hope you can look into #298 or #300 soon. Custom filter function (or called search method) is useful when trying to filter the options according to some other properties of the options (e.g. Chinese pinyin). I'm using my pr #298 in my own project now.

meteorlxy:
@sagalbot Reverted.

Btw, I'm not sure but maybe we could wrap the `spinner` slot like this? So we don't have to add `class=spinner` and `v-show="mutableLoading"` manually
```js
<div class="spinner" v-show="mutableLoading">
  <slot name="spinner">Loading...</slot>
</div>
```

PR #298, opened 2017-08-26
Combine #295 and #296. Add custom filter function.

meteorlxy:
Combine #295 and #296.
Add custom filter function via prop `filterFunction(option, label, search)`. By setting custom filter function, we can search the options according to other properties of the option Object.


meteorlxy:
#299

PR #319, opened 2017-09-19
Adding diacritics conversion functionality

adrichey:
Added the ability to remove diacritic marks from a search string, so that search terms do not have to match them exactly. For example, if you search for "Aland Islands" you will now match "Åland Islands". It can be enabled by adding the "remove-diacritics" boolean property to the v-select component.

### DIFF
--- a/dev.html
+++ b/dev.html
@@ -32,6 +32,7 @@
 <body>
     <div id="app">
         <v-select placeholder="default" :options="options"></v-select>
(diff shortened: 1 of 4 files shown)


adrichey:
Totally understandable. You won't hurt my feelings if this doesn't make it in. It had just made my life easier, so I thought I would submit it. 😄

Issue #284, opened 2017-08-12
Strange behavior after size CSS size change.


nicolasvahidzein:
Hello,

after changing a few parameters i when i change the CSS the text box changes height to accomodate for the height of the element in pink which is:

```
.v-select.vselectmerlinaddresscountry input[type=search],
.v-select.vselectmerlinaddresscountry input[type=search]:focus
```

![001](https://user-images.githubusercontent.com/29876108/29241416-314375be-7f71-11e7-848e-bf3f91b37054.jpg)
![002](https://user-images.githubusercontent.com/29876108/29241417-316600b6-7f71-11e7-95b5-c19119129852.jpg)

I went through your module and extracted the complete CSS then customized only the parts i needed. You will find them with this line:

//custom*************************************

but yet there are issues with that block with the input text. Can you please help me?

Here is the complete code:
```
.v-select.vselectmerlinaddresscountry {
    position: relative;
    font-family: sans-serif;
    text-align: center; //custom*************************************
}

.v-select.vselectmerlinaddresscountry,
.v-select.vselectmerlinaddresscountry * {
    box-sizing: border-box;
}

.v-select.vselectmerlinaddresscountry .open-indicator {
    position: absolute;
    bottom: 22px; //custom*************************************
    right: 10px;
    display: inline-block;
    cursor: pointer;
    pointer-events: all;
    transition: all .15s cubic-bezier(1, -.115, .975, .855);
    transition-timing-function: cubic-bezier(1, -.115, .975, .855);
    opacity: 1;
    transition: opacity .1s;
    height: 10px; //custom*************************************
    width: 10px; //custom*************************************
}

.v-select.vselectmerlinaddresscountry .open-indicator:before {
    border-color: rgba(60, 60, 60, .5);
    border-style: solid;
    border-width: 3px 3px 0 0;
    content: "";
    display: inline-block;
    height: 10px; //custom*************************************
    width: 10px; //custom*************************************
    vertical-align: top;
    transform: rotate(133deg);
    transition: all .15s cubic-bezier(1, -.115, .975, .855);
    transition-timing-function: cubic-bezier(1, -.115, .975, .855);
    box-sizing: inherit;
}

.v-select.vselectmerlinaddresscountry.open .open-indicator:before {
    transform: rotate(315deg);
}

.v-select.vselectmerlinaddresscountry.loading .open-indicator {
    opacity: 0;
}

.v-select.vselectmerlinaddresscountry.open .open-indicator {
[...]

PR #346, opened 2017-10-19
Solution to line breaking on short width

setefocos:
Proposal to solve the problem documented at the Issue "Strange behavior after size CSS size change. #284"


MarZab:
this breaks rtl, a solution would be `:style="(dir==='rtl')?'float:right':'float:left'"` 
but we should probably check with someone who uses rtl.

setefocos:
@MarZab I don't use rlt on vue-select.
=/

stevenharman:
Perhaps this is solved by #512? @setefocos, can you confirm?

Issue #210, opened 2017-05-11
You can still delete selected tags when dropdown is disabled


dynamiccarrots:
When I disable the drop down, I expect all of the selected tags to remain as they are but not be deletable.

```
<v-select :on-change="map_allowed_geos" :value="{}" v-model="allowed_geos" :options="options.geos" multiple :no-drop="true" disabled="true"></v-select>
```

I have tried adding the `disabled` and `:no-drop` props but I am not getting the desired result.

Is there another prop that I am missing?

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [typeahead keypress handlers are still active when dropdown is disabled](https://trello.com/c/9avkuZYL/23-typeahead-keypress-handlers-are-still-active-when-dropdown-is-disabled)

sagalbot:
Further to that, you can actually 'navigate' the dropdown with the up and down arrows and make new selections.

Issue #336, opened 2017-10-10
bug: disabled not completely disabled (item can still be removed)


akhchan99:
With the newly added disabled option, the drop down box is disabled, but when it is used with multiple selection, the selection can still be removed

https://codepen.io/akhchan99/pen/EwRayP

* moving the mouse over the drop down box shows disabled icon
* but when moved to the selected item's 'X' (remove button), the remove function still works

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [typeahead keypress handlers are still active when dropdown is disabled](https://trello.com/c/9avkuZYL/23-typeahead-keypress-handlers-are-still-active-when-dropdown-is-disabled)

PR #348, opened 2017-10-20
Prevent interaction when component is disabled

andywarren86:
- can no longer remove selected items from disabled multiple select by mouse click
- can no longer interact with dropdown via keyboard


andywarren86:
Fixes issues:
https://github.com/sagalbot/vue-select/issues/210
https://github.com/sagalbot/vue-select/issues/336

ThomHurks:
I was actually thinking about making a PR for this too, but you beat me to it. Nice!

andywarren86:
I had considered removing the close button, but was worried it would cause all of the items to change size and position if the disabled prop was dynamically toggled. Not sure if that is worth worrying about though?

ThomHurks:
I think it's better to leave the remove button for accessibility reasons too. The general practice seems to be to not remove disabled items, so that the impaired user still knows they exist, but through proper use of ARIA attributes knows that the button is disabled and (potentially) for what reason. For regular users it's probably better too to see that the labels are potentially removable, but that the form is just disabled.

andywarren86:
No worries glad I could help!

Issue #349, opened 2017-10-20
Coloured options/labels (gathering feedback before PR)


ThomHurks:
I'm thinking about making a PR for giving options a color. You might have labels or tags with certain colors assigned, perhaps a unique color per label/tag, or perhaps a color per group.

Now, there are various ways to implement this, so before I try anything I would like to gather some opinions:
- You could potentially pass a color along with the data for each label, and most likely tell v-select what data attribute to look for the color in.
- However, this is quite strict. You can also just let the developer pass in an (array of) class(es), and then you add the classes on the labels. As before, you can tell v-select in what attribute to look for the class(es). This is a more generic approach. Note that this is different from creating a theme, this just influences styling for specific labels that the dev wants to look different.
- You could, depending on what API you like, even pass in label-class pairs to v-select separately from the data.
- Some other things to think about: do you only show the color after you add a label? Or do you already make the label text coloured when a user is looking at the whole list of options? For accessibility it is probably better to just let the list of options be black text on white background instead of getting rainbox text, but perhaps there is another better way.

What is your opinion? I would like this functionality, so based on a certain preference I can write some code :)

MarZab:
I would suggest making use of slots (make the existing slot 1 level higher).

ThomHurks:
@MarZab I tried implementing it like that, but you lose the delete button that way since it's a sibling of the current slot. You move the slot one level higher you can customise how the selected tag should look, but you lose the fallback content of the slot and thus the delete button. Do you have any idea to solve that?

MarZab:
See the commit above.

The downside is a extra <span> that could possibly be migitated.

ThomHurks:
Thanks @MarZab! I based your commit off sagalbot/master and created the PR. Seems to work nicely for my use-case and all tests pass.

sagalbot:
@ThomHurks sorry for the delay on this one! I had a head/neck injury a couple weeks back and have been taking it super slow. Thanks for raising this, hoping to dive into this one this week!

ThomHurks:
No worries, hope you recover quickly!

julienmonty:
Not sure if the "unique color per label/tag" feature has been implemented in #420. Cannot find anything about this in docs or PR description.

MarZab:
It was, its in the master branch - but it's not in `dist/vue-select.js` and https://www.npmjs.com/package/vue-select also has the old version.

@sagalbot how are you doing? Would you like some help with this project?

[...]

PR #352, opened 2017-10-23
Add a higher level slot to labels so they can be customised completely

ThomHurks:
As discussed in #349, it would be very nice if labels can be customised. My use-case was to give a colour to certain tags to aid visual identification, but a more generic customisation system is probably better to fit everyone's use-case and to not make vue-select too complex.

This PR adds a higher level slot to the labels so that the developer can replace them completely with what they want. So adding a colour to labels is easy, but you could even do something very extravagant if you want to. The already existing slots are not touched to be compatible.

In my case, I replace the label and add a colour in this way:

```HTML
<span slot="selected-option-container"
      slot-scope="props"
      v-bind:key="props.option.index"
      class="selected-tag"
      :style="{'background-color': props.option.name === 'foo' ? 'red' : 'blue'}">
    {{ props.option.name }}
    <button @click="props.deselect(props.option)" type="button" class="close"
            aria-label="Remove option">
        <span aria-hidden="true">&times;</span>
    </button>
</span>
```

Thanks to @MarZab

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -282,14 +282,17 @@
   <div :dir="dir" class="dropdown v-select" :class="dropdownClasses">
     <div ref="toggle" @mousedown.prevent="toggleDropdown" :class="['dropdown-toggle', 'clearfix', {'disabled': disabled}]">
 
-      <span class="selected-tag" v-for="option in valueAsArray" v-bind:key="option.index">
-        <slot name="selected-option" v-bind="option">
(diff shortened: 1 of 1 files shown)


SergeyMiracle:
Waiting for this PR to be merged, will help in my project a lot 👍

ThomHurks:
No worries, thanks for merging!

hannesaasamets:
@sagalbot Can you publish this also?

PR #353, opened 2017-10-24
Replace '.form-control' class with '.v-select-search' class.

owenconti:
`.form-control` clashes with the bootstrap styles, which causes unnecessary CSS hacks to get around.

Anyone who is using `vue-select` and relies on this `.form-control` should be aware of this change during an update.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -307,7 +307,7 @@
               @blur="onSearchBlur"
               @focus="onSearchFocus"
               type="search"
-              class="form-control"
+              class="v-select-search"
               :disabled="disabled"
               :placeholder="searchPlaceholder"
               :readonly="!searchable"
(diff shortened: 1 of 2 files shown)


adi518:
This is not how you solve css conflicts with components, regardless of Vue. To solve this, the author should scope their component's css, and instead of adding bootstrap globally, you should use it in its component form, where its `.form-control`class will remain scoped to a FormControl component. The global thinking is anti-pattern to this kind of development. To solve the customization problem of a component, we have a feature called `scoped-slot`, which allows to take the processed component data and compose our own template. @sagalbot will have to implement it at some point.

jaapio:
@sagalbot think this one can be closed. there will always be collisions between css global and components. If this is an issue the fix isn't the right one.

Issue #356, opened 2017-10-27
[Bug] documentation/example, on-change is a property and must be set via :on-change="callback"


pniederlag:
Hi,

there seems to be a bug in the documentation for the example for the on-change callback

https://github.com/sagalbot/vue-select/blob/f26822cee6a306b65157425c1334483e2b850f4b/docs/components/Examples.vue#L125

```

    // code on example
v-on:change="callback" 
    // should be the same as
@change="callback"  
    // neither one ^^ is working

    // from reading the code and learning vuejs it seems onChange is a property of <v-select/>
:on-change="callback"
    // works :->
```

pniederlag:
https://github.com/sagalbot/vue-select/pull/332 Now seen there as already a better PR

PR #332, opened 2017-10-06
on-change docs fix

skojin:
fix on-change examples in docs

replace `v-on:change` with `:on-change`

PR #357, opened 2017-10-27
[DOC] :on-change property in example

pniederlag:
on-change is a property and must be addressed as :on-change

Fixes: #356


pniederlag:
there is already a better PR that fixes this in another spot as well:

https://github.com/sagalbot/vue-select/pull/332

PR #365, opened 2017-11-02
Change color when .v-select is .selected-tag

ivomarsan:
Change background-color to font-weight: bold


ivomarsan:
I think it would be better to replace the effect when the value is selected from background-color to bold

vitamink:
I would disagree with this because it goes against the expected behavior of select lists. If you want this styling then why not override the CSS in your own stylesheet, specifying `!important` to ensure that your override is respected.

ivomarsan:
OK!

Issue #354, opened 2017-10-25
Close crosshair on single select


orionych:
Please make possible to use a single select crosshair to remove a selected value and set an option to null!

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [add 'x' clear button](https://trello.com/c/rqBZ0ONM/13-add-x-clear-button)

sagalbot:
Already on the roadmap. Might be able to squeeze that into `v2.3.1`.

Issue #211, opened 2017-05-12
No way clear selected item when only one is allowed


ankurk91:
vue-js v2.3.3
vue-select v2.2.0
```html
<!-- multiple selection not allowed -->
<v-select              label="name"
                            :debounce="250"
                            :on-search="getCategories"
                            :on-change="onChangeParent"
                            :options="searchedCategories"
                            placeholder="Search category">
                  </v-select>
```
![screenshot from 2017-05-12 10-28-35](https://cloud.githubusercontent.com/assets/6111524/25982876/c79460e4-36fd-11e7-9b7a-263ca46e0e0f.png)
As you can see in screenshot there is no close button to clear the select box. Once item has been selected there no way to undo.

But if multiple selection are allowed then it shows close button next to each selected item.
![screenshot from 2017-05-12 10-32-46](https://cloud.githubusercontent.com/assets/6111524/25982937/5dcf9b46-36fe-11e7-8887-58b63720b4ea.png)

P.S: Backspace does clear the selection, but it would be nice to have a visible clear button

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [add 'x' clear button](https://trello.com/c/rqBZ0ONM/13-add-x-clear-button)

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [add 'x' clear button](https://trello.com/c/rqBZ0ONM/13-add-x-clear-button)

sagalbot:
Agreed! This is on the roadmap: ![](https://github.trello.services/images/mini-trello-icon.png) [add 'x' clear button](https://trello.com/c/rqBZ0ONM/13-add-x-clear-button). Going to close this for now but I'll let you know when it's complete.

blackboxx:
Agree To, how to clear 👍

PR #368, opened 2017-11-05
Add clear button to single select

andywarren86:
Hi Jeff,

This is a work in progress that adds a clear button to a single select to remove the currently selected value.

Still to do:
- Add a prop to turn this feature on & off (defaults to off?)
- Update docs
- Add some tests
- anything else you can think of?

Can you please take a look over and if you're happy I will complete the PR.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -23,6 +23,10 @@
   .v-select.rtl .dropdown-menu {
     text-align: right;
   }
+  .v-select.rtl .dropdown-toggle .clear {
+    left: 30px;
+    right: auto;
+  }
   /* Open Indicator */
   .v-select .open-indicator {
     position: absolute;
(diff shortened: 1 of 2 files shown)


andywarren86:
References issues #354 #211

andywarren86:
I can appreciate if you don't want to add another prop as there is already quite a few. Changing to a slot sounds like the way to go then as it will provide the most flexibility.

andywarren86:
Sorry for taking so long to get back to this.

I had a go wrapping the button in a slot and while it gives the user the ability to customise the display of the button, they lose the default behaviour attached to v-show and @click. Are users expected to re-implement these methods themselves, or is there some way to structure things so they can still leverage the existing methods?

Regardless I've added some tests so I'm happy if you'd like to merge this now and worry about slots later if somebody requests it. As you've mentioned you can still turn it off easily enough via CSS if necessary.

fdenzer:
👍 Looks perfect, merge this please?!

andywarren86:
Hi @sagalbot, is there anything else you would like me to do for this issue?

iyerusad:
#1383 seeks clear button when in multiple mode.

Am wondering why `showClearButton()` requires `!this.multiple`? Presumably removing that condition would resolve the above feature request, but not clear why its in place so figured would ask.

https://github.com/sagalbot/vue-select/commit/a9ae09edc03285705ed3f05eb7f602f7c55ced76#diff-96b70cc49a38daa29e6746780b2601d0e779a5a13f8b1afaad755aed127d8825R985
```
      showClearButton() {
        return !this.multiple && !this.open && this.mutableValue != null
      }
```

PR #378, opened 2017-11-25
Custom filter function, now with tests

SKalt:
An update of @meteorlxy's  custom filter function PR to address #295 and #296.  I've added unit tests in the hope of this PR getting accepted, plus minor stylistic tweaks (014c51d, 10e8516).  In e30a000  and 57c084d I closed an edge case where valueAsArray might not return an array.


Mark-de-Haan:
+1 Would be really happy to see this being added to the project

PR #380, opened 2017-11-27
Use deep equal in mutableValue

intersrc:
Sometimes, `val` & `old` are objects, like `{ label, value }`, `val !== old` is true, but they have same `label` & `value`, this may cause a dead loop.  
It's better to use a deep equal here.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -594,7 +594,7 @@
         if (this.multiple) {
           this.onChange ? this.onChange(val) : null
         } else {
-          this.onChange && val !== old ? this.onChange(val) : null
+          this.onChange && JSON.stringify(val) !== JSON.stringify(old) ? this.onChange(val) : null
         }
       },
 


intersrc:
The test added. Now use `val !== old` will make tests die.  
There is a small problem: if that happened, we can't locate which test is broken because of the dead loop.

intersrc:
Solved.

src/components/Select.vue
```diff
@@ -594,7 +594,7 @@
         if (this.multiple) {
           this.onChange ? this.onChange(val) : null
         } else {
-          this.onChange && val !== old ? this.onChange(val) : null
+          this.onChange && JSON.stringify(val) !== JSON.stringify(old) ? this.onChange(val) : null
```
Lexxus: This is not a good comparison because order of properties can be different but values can be equal, so as a result comparison will fail.

skjnldsv:
This is really an issue that needs to be fixed :/

src/components/Select.vue
intersrc: Modified.

PR #386, opened 2017-12-06
Add prop to disable option filtering

annismckenzie:
First off, thank you for this library! :)

This came up in one of our projects where the options are already filtered by the term the user types in and the returned items don't necessarily contain that search term (think full-text searching for users by their email/first name/last name/etc. and only returning their name for display). Without an option to disable the options filtering it always shows an empty list.


taylorotwell:
+1 Need this badly. Makes the component hard to use with any pre-filtered results from the server side.

annismckenzie:
Woohoooo, thank you! 💃 

Making a mental note to »allow edits from maintainers« next time and skip the dist file update. ☑️

PR #394, opened 2017-12-14
Allow the createOption function to reject the new option

ph4r05:
I came to a need to block creating of some particular tags so this solution came to my mind.

If user-defined callback `createOption()` returns `null` or `undefined` the option will be rejected / not added.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -679,7 +679,9 @@
             option = this.createOption(option)
           }
 
-          if (this.multiple && !this.mutableValue) {
+          if (!option) {
+            // nothing.
+          } else if (this.multiple && !this.mutableValue) {
             this.mutableValue = [option]
           } else if (this.multiple) {
             this.mutableValue.push(option)


ph4r05:
This appears to break the tests. I will take another look.

eric-naguras:
@ph4r05 May I ask why this PR was not honored?
This issue is a breaking problem.

I want my users to confirm when they add a new item to the list, like this:

```
createComposer (newComposer) {   
      console.log('New Composer', newComposer)   
      if (window.confirm(`Not in list.\n\nDo you want to add ?`)) {   
        return {label: newComposer}   
      } else {   
        return null   
      }   
    }
```
But the current version will crash the component.

This PR solves the problem although I would not have an empty if block.
Can you fix this yourself or come up with a better solution.

Cheers.

tapanbavaliya:
i am facing this problem, can anyone tell me the status of this please

ph4r05:
I was not working on this since closing the PR. There was apparently some problem with breaking tests and I decided to take a different way.

Take a look if my fork works for you: https://github.com/ph4r05/vue-select

I used own component: https://github.com/ph4r05/ph4-vue-select

Issue #292, opened 2017-08-17
Custom Value [v-bind:value]


kJamesy:
In the same fashion as custom labels (http://sagalbot.github.io/vue-select/#ex-labels) how do I set my custom value for when my options is an array of objects?

Toomean:
+

meteorlxy:
I also want this feature. 

The value is the `Object` itself by default. 

If you want to get custom value without modifying `vue-select` itself, you may try to use `vuex` and set the `onChange(val)` prop

htunnicliff:
The following behavior seems to defeat the expected functionality of `v-model`.

## Component
```html
<v-select v-model="countryCode" :options="countriesWithLabels"></v-select>
```
```js
import vueSelect from 'vue-select'

export default {
  data() {
    countryCode: 'US',
    countriesWithLabels: [
      {
        label: 'United States',
        value: 'US'
      },
      {
        label: 'France',
        value: 'FR'
      },
      {
        label: 'United Kingdom',
        value: 'UK'
      }
    ]
  },
  components: {
    'v-select': vueSelect
  }
}
```

-----

## Problems
Loading v-select while `countryCode` is set to a string such as "US" produces the following problems:

1. v-select fails to match "US" to the corresponding option that labels it as "United States", and renders the option as "US".
2. Upon choosing a different option from the dropdown list, **v-select updates the value of the v-model** (`countryCode`) **to _the option object itself_, rather than `option.value`**.

```js
// Expected result when selecting "United Kingdom" option
data() {
  ...
  countryCode: 'UK',
  ...
}

// Actual result
data() {
  ...
  countryCode: {
    label: 'United Kingdom',
    value: 'UK'
  },
  ...
}
```

lucini:
Same here, any solution?

vitamink:
The root of this behavior seems to be in the **getOptionLabel** function. I've made a rough fix, but cannot be sure that it'll behave in every possible scenario. Works for me though :-)

      getOptionLabel: {
        type: Function,
        default(option) {
          if (typeof option === 'object') {
            if (this.label && option[this.label]) {
              return option[this.label]
            }
          } else if (typeof option === 'string') {
            let label = null
            this.options.forEach(opt => {
              if (typeof opt === 'object' && opt[this.value] === option[this.label]) {
                label = opt[this.label]
              } else if (typeof opt === 'object' && opt[this.value] === option) {
                label = opt[this.label]
              }
            });
            if (label != null) return label
[...]

PR #398, opened 2017-12-18
Initial fix for issue 292.

vitamink:
Fixes issue #292, specifically as described by @htunnicliff.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -690,11 +690,24 @@
           }
 
           if (this.multiple && !this.mutableValue) {
+            if (typeof option === 'object' && option.value) {
+              this.mutableValue = [option.value]
+            } else {
+              this.mutableValue = [option]
+            }
             this.mutableValue = [option]
           } else if (this.multiple) {
-            this.mutableValue.push(option)
+            if (typeof option === 'object' && option.value) {
+              this.mutableValue.push(option.value)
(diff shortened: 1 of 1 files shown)


vitamink:
Please ignore this while I investigate an issue.

vitamink:
Hello again.

It's been tested with single and multi select.
I've also tested it with simple (string) value arrays, and `{value: 'foo', label: 'bar'}` arrays.
Lastly, I tested it with different label names such as: `{value: 'foo', myLabel: 'bar'}`, with the Label= override, and that works.

I hope this is all good. I'd be happy to hear any comments/bugs. Otherwise, I think this can be merged without issue.

Thanks.

tspesh:
Any updates on this??

vitamink:
I recognise that this is effectively a breaking change. To my mind the existing behavior of this component is incorrect, in that the whole point of the `value` property is that it contain a value, and not the object from which the value was derived.

@sagalbot Perhaps this needs some discussion? I may have missed the point entirely, and there is a very easy way to achieve the desired behavior.

vitamink:
Yes, I noticed the existence of `label`, but not `value`.

Would you be happier if a `value` property was added to the component, and if null, then the current behavior would apply. If not null, then the object[value] would be used.

That would make it optional, and not break any existing code...

PR #194, opened 2017-04-25
Add options to enable/disable deselect when repeat select an option

msyesyan:
https://github.com/sagalbot/vue-select/issues/180

as the issues I said, I just want to keep the selected option when I repeat select the same options

toggleselectoption

PR #436, opened 2018-01-23
dont deselect when clicking selected options

### LINKED ISSUE #180: select same item will set its value to null
I see in your documents demo, when you repeat select the same value, the option will not be deselected, but when I use this plugin, it will deselect this same option.

I see in your source code there is a logic `if (this.isOptionSelected(option)) this.deselect(option)`. 
I don't want this behavior, I just want to keep the selected option if user select the same option in the dropdown list

sagalbot:
Changes behavior so that clicking an item in the dropdown is only for selection - it will never deselect an item. `multiple` selects have clear buttons for each tag, and `single` selects have the clear button.

The html `<select>` does not deselect values when clicking the same option, `vue-select` shouldn't either.

---

> Thanks to @msyesyan for filing the issue and the original PR #194. I would have merged that branch now, but there were a lot of conflicts and I wanted to get this done quickly after holding off for so long. 

Closes #180 
Closes #194 (duplicate)

### DIFF
--- a/dev/dev.html
+++ b/dev/dev.html
@@ -7,7 +7,7 @@
     <!--<link href="https://cdnjs.cloudflare.com/ajax/libs/foundation/6.3.1/css/foundation.min.css" rel="stylesheet">-->
     <!--<link href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.3.2/css/bulma.min.css" rel="stylesheet">-->
     <!--<link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" rel="stylesheet">-->
-    <!-- <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.0.0-alpha.6/css/bootstrap.min.css"> -->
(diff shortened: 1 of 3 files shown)


enterlight:
Hello!  I am sorry if I am wasting your time here by doing something wrong...  I just downloaded commit #e35e2ac100c355352be543c82201474b949c53ef and it does not seem to work... If there is an item selected and I select that same item, then it gets deselected...  I checked and rechecked the code to make sure npm installed the right version and yet it does not work...

codynguyen:
Hi @sagalbot 

> The html <select> does not deselect values when clicking the same option, vue-select shouldn't either

Sorry to bring this back, but I think it's a pretty strong opinion. Maybe it's better to turn on/off the deselect for it to fit more use cases / designs? Not every one wants the same behavior as the html `<select>`

https://github.com/sagalbot/vue-select/pull/1033#discussion_r496702280

codynguyen:
Hi @sagalbot

Sorry to bring this back, but maybe it's better to turn on/off the deselect for it to fit more use cases / designs? Not every one wants the same behavior as the html `<select>`

https://github.com/sagalbot/vue-select/pull/1033#discussion_r496702280

PR #438, opened 2018-01-23
Disable browser autocomplete FIX

enid-colic:
The autocomplete attribute value `"false"` does not exists there are only two options `"on"` or `"off"`
**Syntax:** `<input autocomplete="on|off">`

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -336,7 +336,7 @@
               @focus="onSearchFocus"
               type="search"
               class="form-control"
-              autocomplete="false"
+              autocomplete="off"
               :disabled="disabled"
               :placeholder="searchPlaceholder"
               :tabindex="tabindex"


denone:
@enid-colic thanks for fixing it, but can you please release this fix as well? 
Also, take a look here
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

Quite interesting thing is written at the end. Basically, according to last paragraphs, setting to "false" might also right thing 😃 

> In some cases, the browser will continue suggesting autocompletion values even if the autocomplete attribute is set to off. This unexpected behavior can be quite puzzling for developers. The trick to really enforcing non-autocompletion is to assign an invalid value to the attribute, for example:

> `autocomplete="nope"`
> Since this value is not a valid one for the autocomplete attribute, the browser has no way to match it, and stops trying to autocomplete the field.

Issue #430, opened 2018-01-20
Display bug on narrow select width


bzarzuela:
![image](https://user-images.githubusercontent.com/230894/35181884-85b62cdc-fe05-11e7-995d-baeb01319305.png)

I believe this display issue is being caused by the input element. See:

![image](https://user-images.githubusercontent.com/230894/35181889-9b0196da-fe05-11e7-837e-03f2dee697f3.png)

I looked at the code and found `:style="{ width: isValueEmpty ? '100%' : 'auto' }"` on line 344 of the component. I think the correct fix for this would be to modify the condition to be "is the value empty and is it a multi-select?"

If I set the width to 0px, the issue resolves itself. Please take a look.

![image](https://user-images.githubusercontent.com/230894/35181907-edca779c-fe05-11e7-8503-abd69ba43f23.png)

I'll make a PR if this is the right way to resolve the issue. Please let me know.

maxorlovsky:
In firefox it's even worse.

sagalbot:
Yep, the `input` is the issue. Would gladly accept a PR with a fix, this one has been on the list for too long.

bzarzuela:
Thanks! I opened PR #441

ModestasV:
Any luck this could get merged? We have the same issue and seems like there is a fix but hasn't been merged :)

egistli:
As the side effect shown in #441, it seems another approach is to make selected tags of `dropdown-toggle` to be horizontally scrollable and leave the input width as is and being the left most item in the `dropdown-toggle`. The effect is like following:

![screen shot 2018-02-14 at 8 57 40 am](https://user-images.githubusercontent.com/112525/36182556-34704344-1165-11e8-9213-2ea88b567862.png)

Not sure what your thoughts on this though since there might be more use cases than I considered and it's also an aesthetic problem.

fractalocity:
Did this ever get resolved or find a css fix for this? I tried the css fixes mentioned above but it didn't work. When i adjust the 'width', this adjusts the width of the v-select, not the input object. I checked out #430 but am not seeing a fix. 

I was hoping I could modify the width in the <style></style> tag but can't (I think because it gets overwritten pragmatically)
![image](https://user-images.githubusercontent.com/29801289/36316352-b26d3b44-12ff-11e8-8c4d-705dbbe685d0.png)

nahuelhds:
In my case, using  a single select I've wrapper the v-select in a container div and styled it like this

```
.v-select-container {
[...]

PR #441, opened 2018-01-28
Fix for #430 Narrow Select Width Display Bug

bzarzuela:
Fixes a display bug I encountered in my project where I had long options on a narrow select width. With this, the premature wrapping to the next line as well as the height restricted silver background on a selected tag is fixed.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -341,7 +341,7 @@
               :placeholder="searchPlaceholder"
               :tabindex="tabindex"
               :readonly="!searchable"
-              :style="{ width: isValueEmpty ? '100%' : 'auto' }"
+              :style="{ width: isValueEmpty ? '100%' : '1px' }"
               :id="inputId"
               aria-label="Search for option"
       >
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -141,7 +141,7 @@
     background-color: #f0f0f0;
     border: 1px solid #ccc;
     border-radius: 4px;
-    height: 26px;
+    min-height: 26px;
     margin: 4px 1px 0px 3px;
     padding: 1px 0.25em;
     float: left;


bzarzuela:
I see. Okay, I’m going to work on these side-effects. I’ll take that
approach and have the input automatically resize to the width of the text.
I think a computed property would do it.

pippinsoft:
Maybe this? `:style="{ width: isValueEmpty ? '100%' : (multiple ? 'auto' : '1px') }"`

mkudenko:
@sagalbot Can this be merged please?

pippinsoft:
This seems to work: `:style="{ width: isValueEmpty || search ? '100%' : '1px' }"`

fractalocity:
It looks like this got merged, but how do you get this to work. I tried  `:style="{ width: isValueEmpty || search ? '100%' : '1px' }"` and that didn't seem to do anything. I tried just forcing the style width to 100% or 1px also and that didn't seem to make a difference.

fractalocity:
@pippinsof are you sure that style modification worked for you? It doesn't not work for me.
`<v-select :style="{ width: isValueEmpty || search ? '100%' : '1px' }" ... >`

pippinsoft:
@fractalocity It does work for me, although I have quite a few modifications, so it could be that I missed something. Are you changing the file under `node_modules`? That won't work. You need to copy it to your source (or explicitly include the `Select.vue` file) and hard refresh to get see changes in your browser.

pippinsoft:
@fractalocity Sorry, I think I see what's causing the confusion. My comment is in reference to this line:

https://github.com/sagalbot/vue-select/blob/546b27d7d21e77680b47efe50d8873237f7a775f/src/components/Select.vue#L344

Which is changed in the PR.

fractalocity:
@pippinsoft You are correct. I was hoping I could modify the input object's style but using the v-select ':style=' property.

I gave your suggestion another shot and still failing. 
This probably won't be a great question, but what do you mean by 'copy it to your source, or explicitly include the Select.vue file)'? 

I tried adding the Select.vue to my project directory (project/src/components) and then added the mixins folder that Select.vue needed to my src directory (project/src). When I did this I got a lot of error messages so that is probably not what you meant in your suggestion:
image

I also tried importing the Select.vue file within the nodes_mode directory which also did not work:
image

fractalocity:
@bzarzuela were you ever able to get the input width to function like you wanted it to?

bzarzuela:
@fractalocity yes, but I haven't found any time yet to modify it so that it would address the issues @sagalbot raised. I will try to do so next weekend.

pippinsoft:
@fractalocity Looks like you're on the right track. Something's upset about tabs -- try swapping them for spaces. (I saw another PR that had this change.)

fractalocity:
@pippinsoft thanks for the guidance. I'll tinker with it a bit more

fractalocity:
@bzarzuela thanks for the reply. I can try tinkering around with it until then.

eriknygren:
I believe I fixed the issue here without raising any side effects on the multiple select mode, would appreciate your feedback https://github.com/sagalbot/vue-select/pull/512 @fractalocity @bzarzuela, cheers.

xavism:
@fractalocity you can do it this way, but you should import the index.js inside the src folder. There the component is exported to use 😄 

image

stevenharman:
I think we can close this one; it’s been superseded by #512.

Issue #460, opened 2018-02-14
The 'current selection' area is not displaying nicely [UI]


fractalocity:
Hi, the vue-select component is really nice. Does a great job functionality wise. I'm having a bit of frustration with the UI.

The 'current selection' area has an additional line of whitespace. my selection can only take up less than half of the selection text box ... if the text is too long ... i guess wrapping is turned on and an unneeded line of white space appears in the 'current selection' area. I've tried setting padding to 0 hoping that would give more room, but this trick didn't really help. It would be really like to use the tool; it's really cool, but the 'current selection area' for the UI is a bit wonky.

Please see the pictures below in case there is any misunderstanding as to which white space this message is about.

I reviewed other issues and i'm guessing this tool worked fine before some people put in feature requests for wrapping and other (#241, #176, #340, #308). 

I would sort of consider this UI behavior a bug. 

If there is a property which I can set which prevents this from happening which I'm overlooking, that would be awesome. 

Figure 1: Ideal ... no white space but the select needs to be insanely long for a small bit of text
![vue_select_normal](https://user-images.githubusercontent.com/29801289/36229747-eaa31ac8-119d-11e8-88af-4365bb0ec8f1.JPG)

Figure 2: This is what happens when i start to shrink the internet browser, thereby also shrinking the width of the select
![vue_select_when_scrunched](https://user-images.githubusercontent.com/29801289/36229751-ee1250b6-119d-11e8-9620-6a824d4bd1f6.JPG)

Figure 3: The below pictures show that same thing happens on the vue-select website
![vue_select_question](https://user-images.githubusercontent.com/29801289/36229577-5ad03dae-119d-11e8-8b63-3438e4ccd825.JPG)

Figure 4: shows same as figure 3
![vue_select_question_2](https://user-images.githubusercontent.com/29801289/36229581-5e5e4b00-119d-11e8-98cb-4fc31457b026.JPG)

nahuelhds:
Hi. I'm using 2.5.1 and I'm having issues with this. See next image, I've just added more chars to the values from the Docs

![image](https://user-images.githubusercontent.com/1885647/47237874-b50f7d80-d3b6-11e8-8dee-035e9237206c.png)

abtv:
@nahuelhds have you fixed it?

wmarcos:
Doesn't work for me in v3.1.0

I did an ugly workaround changing css 'display' property when a value is selected listening to 'input' event.

<v-select ref="select" @input="onInput" >
[...]

Issue #306, opened 2017-09-03
Potential build error


gregaz:
When trying to build this project, I get the below error:

```
npm install -g vue
npm install vue-select
cd node-modules/vue-select
npm install
npm run build

ERROR in vue-select.js from UglifyJs
SyntaxError: Unexpected token: name (VueSelect) [./src/index.js:1,0]
```

This makes me think that vue isn't installed correctly, but I have tried various ways of installing it via npm (installing it globally, installing it as a dependency instead of a peer dependency, installing it locally).

I wouldn't be surprised if this was being caused by something I was doing wrong, I'm new to js/npm/webpack. Any help would be greatly appreciated.

sagalbot:
No need to run the build to use it in your project. Running the build is only done when a new version of vue-select is ready to release.

If you're using a module bundler (webpack), you just need to include the modules to use them:

```js
import Vue from 'vue';
import VueSelect from 'vue-select';
```

gregaz:
I want to fork vue-select for my project, which will require me being able to build it, right?

sagalbot:
Yes it will. In that case:

```
$ git clone git@github.com:gregaz/vue-select.git 
$ cd vue-select
$ npm install
$ npm run build
```

gregaz:
That worked for me! It's a matter of cloning the repo vs installing it via npm. Thanks!

andreiciceu:
Hello,

I forking & cloning, but stumbled upon the same problem:
```
Hash: 97ccbd3f7cb7f9d8a564
Version: webpack 1.15.0
Time: 52ms
            Asset     Size  Chunks             Chunk Names
    vue-select.js  2.04 kB       0  [emitted]  lib
vue-select.js.map  2.31 kB       0  [emitted]  lib

ERROR in vue-select.js from UglifyJs
SyntaxError: Unexpected token: name (VueSelect) [./src/index.js:1,0]
```
How can I solve it?
Many thanks

CraftLogan:
I am having the same issue and I cloned but still get this error 

ERROR in vue-select.js from UglifyJs
SyntaxError: Unexpected token: name (VueSelect) [./src/index.js:1,0]


Any help would be appreciated. Thanks

Issue #225, opened 2017-05-29
v-select inside bootstrap coumn gives bad result


goalie7:
Hello!

Whenever I put the v-select component inside a bootstrap col-*-* I get the following result:

![image](https://cloud.githubusercontent.com/assets/7350351/26548489/bea8f1fa-4474-11e7-817d-3e701ebd7b66.png)

Any ideas on how to solve it? Anyone experiences the same?

Regards

goalie7:
Okai it seems that is something related with the select width. If I set the label tol col-md-4 and the v-select to col-md-8 is working as expected.

![image](https://cloud.githubusercontent.com/assets/7350351/26548695/bd084304-4475-11e7-8350-fd44056e255f.png)

Anyways what if I dont want the v-select component to be so large?

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [Text input requires too much horizontal space](https://trello.com/c/CxmRcuYr/24-text-input-requires-too-much-horizontal-space)

stefan-angelov:
I had the same problem and solved like this :
`.v-select {
        font-size: 0.875rem;
        .dropdown {
          .dropdown-menu {
            min-width: initial;
          }
        }
        .dropdown-toggle {
          display: flex;
          flex-wrap: wrap;
          .selected-tag {
            position: relative;
            padding: 0px 15px 0px 5px;
            margin: 2px 1px 2px 3px;
            height: auto;
            button {
              position: absolute;
              right: 2px;
              top: 50%;
              transform: translateY(-50%);
              span {
                vertical-align: super;
              }
            }
          }

          input[style] {
            flex-basis: 20px;
            flex-grow: 1;
            height: 33px;
            padding: 0 20px 0 10px;
            width: 100% !important;
          }
        }
      }`

sagalbot:
@stefan-angelov would you like to do a PR with that fix?

camilosw:
Any hope to see this solved?

MohammadSabbagh:
@stefan-angelov  solution is working 

Thanks

niti11:
Hi All
How can we add the css to this select box.
I want to remove the border for select box & also want to set postion as top for the option box.

I tried this way.
.v-select .dropdown-toggle
{
	border: none !important;
}

But It doesn't work for me. Can you help me on this?

stayacid:
worked for me:
<pre>
.vs__selected {
  +input {
[...]

PR #512, opened 2018-04-18
Fixing unexpected linebreak on single selects

eriknygren:
Fixes https://github.com/sagalbot/vue-select/issues/460

What
---
- Hiding the search input field if the component is in the single value
option.
- Making the search input field full width if no options are selected in
	either single or multi select mode.
- Shrinking it to width auto if there are selected entries in multi
	mode.

Why
---

Before:
image

After:
image

Reason for being broken:
image

The component broke into two lines when selecting a a value in single
mode, because an empty, non-interactable input field was pushed down to
the next row if the selected entry had a long label.

PS not sure if I did something weird with the indentation in the spec file?

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -215,7 +215,8 @@
     cursor: pointer;
   }
   .v-select input[type="search"].hidden {
-    display: none;
+    width: 0px;
+    padding: 0;
   }
   .v-select input[type="search"].shrunk {
     width: auto;
@@ -989,7 +990,7 @@
        */
       inputClasses() {
         return {
-          hidden: !this.multiple && !this.isValueEmpty,
+          hidden: !this.multiple && !this.isValueEmpty && !this.dropdownOpen,
           shrunk: this.multiple && !this.isValueEmpty,
           empty: this.isValueEmpty,
         }
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -73,6 +73,7 @@
     padding: 0;
     background: none;
     border: 1px solid rgba(60, 60, 60, .26);
+    min-height: 36px;
     border-radius: 4px;
     white-space: normal;
   }
@@ -213,6 +214,16 @@
   .v-select.unsearchable input[type="search"]:hover {
     cursor: pointer;
   }
+  .v-select input[type="search"].hidden {
+    display: none;
+  }
+  .v-select input[type="search"].shrunk {
+    width: auto;
+  }
+  .v-select input[type="search"].empty {
+    width: 100%;
+  }
+
     /* List Items */
   .v-select li {
     line-height: 1.42857143; /* Normalize line height */
@@ -336,12 +347,12 @@
               @focus="onSearchFocus"
               type="search"
               class="form-control"
+              :class="inputClasses"
               autocomplete="off"
               :disabled="disabled"
               :placeholder="searchPlaceholder"
               :tabindex="tabindex"
               :readonly="!searchable"
-              :style="{ width: isValueEmpty ? '100%' : 'auto' }"
               :id="inputId"
               aria-label="Search for option"
       >
@@ -969,6 +980,18 @@
           loading: this.mutableLoading,
           rtl: this.dir === 'rtl',
           disabled: this.disabled
+        }
+      },
+
+      /**
+       * Classes to be output on input.form-control
+       * @return {Object}
+       */
+      inputClasses() {
+        return {
+          hidden: !this.multiple && !this.isValueEmpty,
+          shrunk: this.multiple && !this.isValueEmpty,
+          empty: this.isValueEmpty,
         }
       },
 
--- a/test/unit/specs/Select.spec.js
+++ b/test/unit/specs/Select.spec.js
@@ -1377,6 +1377,28 @@ describe('Select.vue', () => {
       expect(vm.$children[0].inputClasses.shrunk).toEqual(false)
     })
 
+
+    it('should not apply the "hidden" class to the search input when a value is present, and the dropdown is open', () => {
+      const vm = new Vue({
+        template: '<div><v-select ref="select" :options="options" :value="value"></v-select></div>',
+        data: {
+          value: 'one',
+          options: ['one', 'two', 'three'],
+          open: true
+        }
+      }).$mount()
+      vm.$children[0].toggleDropdown({target: vm.$children[0].$refs.search})
+      Vue.nextTick(() => {
+        Vue.nextTick(() => {
+          expect(vm.$children[0].open).toEqual(true)
+          expect(vm.$children[0].inputClasses.hidden).toEqual(false)
+          expect(vm.$children[0].inputClasses.empty).toEqual(false)
+          expect(vm.$children[0].inputClasses.shrunk).toEqual(false)
+          done()
+        })
+      })
+    })
+
 		it ('should not reset the search input on focus lost when clearSearchOnSelect is false', (done) => {
 			const vm = new Vue({
 				template: '<div><v-select ref="select" :options="options" :value="value" :clear-search-on-select="false"></v-select></div>',
--- a/test/unit/specs/Select.spec.js
+++ b/test/unit/specs/Select.spec.js
@@ -249,6 +249,36 @@ describe('Select.vue', () => {
 			expect(vm.$children[0].isOptionSelected('foo')).toEqual(true)
 		}),
 
+    it('applies the "empty" class to the search input when no value is selected', () => {
+      const vm = new Vue({
+        template: '<div><v-select :options="options" multiple v-model="value"></v-select></div>',
+        components: {vSelect},
+        data: {
+          value: null,
+          options: [{label: 'one'}]
+        }
+      }).$mount()
+
+      expect(vm.$children[0].inputClasses.empty).toEqual(true)
+      expect(vm.$children[0].inputClasses.shrunk).toEqual(false)
+      expect(vm.$children[0].inputClasses.hidden).toEqual(false)
+    }),
+
+    it('applies the "shrunk" class to the search input when one or more value is selected', () => {
+      const vm = new Vue({
+        template: '<div><v-select :options="options" multiple v-model="value"></v-select></div>',
+        components: {vSelect},
+        data: {
+          value: [{label: 'one'}],
+          options: [{label: 'one'}]
+        }
+      }).$mount()
+
+      expect(vm.$children[0].inputClasses.shrunk).toEqual(true)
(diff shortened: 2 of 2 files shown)


fractalocity:
It looks awesome! Nice fix @eriknygren

eriknygren:
Great, thanks guys! Let me know if you need anything more from me. Happy to help

BasConijn:
I think this is the same problem as #225.

eriknygren:
bump, would love to go back to pointing to the master branch on my project 😎 

@sagalbot any idea when you'll have some time to test this? anything I can do to help?

eriknygren:
@sagalbot can't see any regressions regarding `multiple` select mode, the input used to be `width: auto` during `multiple` select mode, and it remains so.

may-14-2018 18-57-18

Attached a quick video of how it interacts in multiple mode above, looks good to me but you know the component better than me!

aabraham88:
@sagalbot first of all, I really appreciate this beautiful plugin, congrats for your work, it's very useful. I'm using it in an NGO project. Right now I'm experiencing this issue as the other ones commented above. I understand that is just a cosmetic issue, but It would be helpful if you guys can review this pull request and merge into master. Thanks for all. 👍

tursics:
bump ;) Merging this pull request will improve single selections a lot. Leave it as current state would be odd

eriknygren:
@tursics @aabraham88 @Lovatt as a temporary workaround you could always update your dependency to point to this PR until it's merged. That's what I'm doing.

So in your package.json go:
```
    "vue-select": "https://github.com/eriknygren/vue-select.git#3aec695",
```
Clearly we all prefer not being on a fork, but just thought I'd mention it.

Lovatt:
Thanks @eriknygren, i didn't know you could do that!  :relieved:

Lovatt:
@eriknygren i tried your branch in my project, i did indeed fix the formatting issues, but it looks like it breaks search

eriknygren:
> @eriknygren i tried your branch in my project, it did indeed fix the formatting issues, but it looks like it breaks search

@Lovatt hmm, in what way? I'm not convinced that's related to this change. But I'd like to confirm, thanks.

stevenharman:
Just discovered the same issue. Any hope for getting this merged 🔜?

src/components/Select.vue
```diff
@@ -213,6 +214,16 @@
   .v-select.unsearchable input[type="search"]:hover {
     cursor: pointer;
   }
+  .v-select input[type="search"].hidden {
+    display: none;
```
stevenharman: In my testing, `display: none` can mess up the vertical alignment of the clear button and the dropdown arrow. Perhaps we consider the following instead, as they preserve the height of the input element, and the vertical alignment of the other elements?

```css
width: 0px;
padding: 0;
```

stevenharman:
Just discovered the same issue. Any hope for getting this merged 🔜?

For what it's worth, the following hack seems to work around the issue... for now. Maybe?

```css
.v-select.single .selected-tag + input[type="search"] {
  padding: 0;
  width: 0px !important;
}
```

stevenharman:
@sagalbot Sure. You can actually see it in action on the demo pages: https://codepen.io/sagalbot/pen/qxbPLY

Go there in Firefox/Chrome, and then turn on the responsive design dev tool. Shrink the viewport down to less than ~380px wide and you'll see the following:

vue_select_-_validation_with_veevalidate

src/components/Select.vue
eriknygren: Good shout @stevenharman , will update

eriknygren:
@sagalbot just noticed an issue where the input was incorrectly hidden if the user re-opens the dropdown after already having a selected item. I added another commit to fix that, hopefully it’s fairly clear what’s going on. Let me know if you want me to squish the commits or something.

Also, added your suggestion @stevenharman.

https://github.com/sagalbot/vue-select/pull/512/commits/16b6bd7d09708b8637125571e46d28031478f83c

stevenharman:
Nice catch @sagalbot. Looking forward to this being merged and released! Thanks!

stevenharman:
Or, rather... nice catch, @eriknygren. And @sagalbot is that sufficient to reproduce, test the fix, and get this merged and released?

Thanks!

eriknygren:
bump @sagalbot

stevenharman:
@sagalbot @eriknygren The more I play with these changes, the more I'm certain that we should also be "hiding" the input for a multi-select, just as we do for single select. That is, remove the check for `!this.multiple` from the `hidden` class:

```javascript
inputClasses() {
  return {
    hidden: !this.isValueEmpty && !this.dropdownOpen,
    shrunk: this.multiple && !this.isValueEmpty,
    empty: this.isValueEmpty,
  }
}
```

Everything still seems to work as expected, with the added benefit of not adding a second, blank, line to multi-selects.

Possibly related - one way to fix this would be to use a flex-box layout for the selected options, input, loading, and clear buttons.

So two questions:

1. Can we also hide for multi-select?
2 @sagalbot would you be open to using a flex-box based layout for the internals?

eriknygren:
Thanks @stevenharman I'll add your suggestion and try it out later today

eriknygren:
@stevenharman I took your suggestion for a spin, it fixes the bug you've mentioned, and works in all but one scenario. Because the multiple select leaves the input on loosing focus the behaviour is a bit "out of the norm" when this occurs.

**Hopefully this gif explains what I mean:**


multiple_bug


I'd vote for leaving multi select to another pull request. I think this can be merged as-is and it'll add value. Then we can think of another solution for the multiple mode.

stevenharman:
@eriknygren Sounds good. I feel like much of this can be simplified and solved by #594, which I built with your changes in this PR in mind. Once this PR is merged, I'll revisit #594 and tweak it accordingly.

eriknygren:
Perfect, thanks @stevenharman. Let's get this show on the road 🛣

aabraham88:
Excellent work guys! 🎊 Do you know when this will come available in a release? Thanks in advance.

eriknygren:
My pleasure, thanks for taking the time and writing this mega useful component!

aabraham88:
Hi everyone. Sorry for bother you with this question, but how do I do to update the package with npm to have this fix? Thanks in advance.

kyrpas:
@aabraham88 I just managed to add this fix to my project. Here's what I did, I hope it works out for you too.

Since the changes have been merged to master, we could pull the master branch using `npm install --save sagalbot/vue-select#pull/512/head` but it still won't work because the changes are only in the src folder and when you're importing vue-select in your project it pulls the files from the dist folder. So you need to build the dist files again.

According to #306 if you want to rebuild the dist folder, you need to clone the repo in a separate folder. It won't work if you try to npm run install and build inside the vue-select folder in the node_modules of your project.

So, first, I cloned the project into a local directory: `git clone https://github.com/sagalbot/vue-select.git`
Then went into the folder and ran `npm install`. 
Finally I ran `npm run build`.

That built the project and re-created the dist folder using the latest code from `src`.

I then copied over the files from the new dist folder over the node_modules/vue-select/dist folder in my project and everything worked fine. 

This really doesn't feel like the right way to do it but until there's an official release this works fine for me.

aabraham88:
Hi @kyrpas I will try your suggestion. Thanks!

eriknygren:
@aabraham88 @kyrpas Did you try what I previously suggested?
I rebuilt the dist on my master branch, so in your `package.json`:
```
"vue-select": "https://github.com/eriknygren/vue-select.git#3aec695",
```

With you current solution @kyrpas you'll loose the fix if you re-download the dependencies

kyrpas:
@eriknygren I did see your comment on June 4, then realised your fork is a few commits behind the original master branch and chose to rebuild from master. To be honest, none of the two solutions are ideal but since they're only temporary I'm ok with either one. I see your point though. Your suggestion is more robust. Any idea as to when to expect a new release? Is there anything we can do to help?

cor:
I'm also very curious on a release date of a version that includes this patch. When can we expect this? I'll use @eriknygren 's fix in the meantime.

aabraham88:
Hi everyone. I had the opportunity to test this in my project, and I think I found a bug. Here is the gif video of the test. 
http://recordit.co/bUl3YnLj9Q
In my case, I'm using the plugin with Ajax to retrieve the data. The selected value appears there because is retrieved from the database when the component is initialized. The example show the scenario when the user wants to change the current value. Instead of clicking on the "X", I clicked on the Down Arrow, showing me that there isn't any option available. That's because the user didn't write down anything yet.
The expected behavior would be to let the user close the "empty list" and let him write another word replacing the current value.
The problem is that I can't close the dropdown, in order to see the "X" again, and there is no turning back in this state.
I copy here my component props configuration:
```
<v-select
        :options="dataCoordinadores"
        label="nombre"
        placeholder="Escribe el nombre o apellido"
        name="coordinador"
        id="coordinador"
        v-model="coordinadorSeleccionado"
        v-bind:disabled="this.readonly"
        :filterable=false
        @search="onSearch"
        :onChange=this.actualizarCoordinador()
>
</v-select>
```
I hope you can tell me if I'm doing something wrong or instead is an issue. I just want to say that this issue is not reproducible before the change introduced in this pull request.

Thanks!

aabraham88:
@stevenharman no, in my case it didn't work. As you can see in the video, clicking on the  `^` doesn't close the list. I could check with another select input that has preloaded values, that this does not occur. It allows you to click on `^` and close the dropdown.

stevenharman:
Ahh, I just realized I also have #594 applied to my app. That outstanding PR is built atop this one, and has a much more versatile and robust layout scheme. I wonder... if you try that branch, do you still see the issue?

matneves:
Hi there, I'm still seen the unexpected linebreak on single selects.

Captura de Tela 2019-05-21 às 20 51 58

I'm currently using vue-select v3.1.0 along with bootstrap v4.

matt-snider:
@matneves I had the same issue with vue-select v3.1.0 and bootstrap v4. This did it for me in a component that is using vue-select:
 
```vue
<template>
    <v-select
        class="my-select"
        //... other options
    />
</template>

<style scoped>
.my-select >>> input.vs__search {
    height: 0 !important;
    margin: 0 !important;
    padding: 0 !important;
}
</style>
```

Note, I'm using [vue-loader](https://vue-loader.vuejs.org) and the `>>>` is the [deep selector syntax](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors).

wilderdyd:
Hi, The same issue still persists in vue-select 3.10.3. I could solve this for my project with the next CSS properties:

`.v-select .vs__selected-options{
  flex-wrap: nowrap;
  white-space: nowrap;
  overflow: hidden;
}`

PR #538, opened 2018-05-22
fix clickable label

zhilyaev:
# before
Not clickable span

# after 
clickable
clickable-animation

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -312,11 +312,15 @@
   <div :dir="dir" class="dropdown v-select" :class="dropdownClasses">
     <div ref="toggle" @mousedown.prevent="toggleDropdown" :class="['dropdown-toggle', 'clearfix']">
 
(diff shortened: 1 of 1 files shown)


src/components/Select.vue
```diff
@@ -312,11 +312,15 @@
...
+      <slot v-for="option in valueAsArray"
+            name="selected-option-container"
+            :option="(typeof option === 'object')?option:{[label]: option}"
+            :deselect="deselect"
+            :multiple="multiple"
+            :disabled="disabled">
         <span class="selected-tag" v-bind:key="option.index">
           <slot name="selected-option" v-bind="(typeof option === 'object')?option:{[label]: option}">
-            {{ getOptionLabel(option) }}
+            <label :for="inputId" @click="$refs.search.focus()">{{ getOptionLabel(option) }}</label>
```
stevenharman: Wouldn't this mean that if you had custom content in the slot, clicking the labels wouldn't work?

PR #547, opened 2018-05-30
feat(clear) added slot to override clear button

clementdevos:
Added a "clear" slot to override the "x" button with something else (fontawesome icons for instance)

Added an example in the dev file.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -90,13 +90,17 @@
     position: absolute;
     bottom: 9px;
     right: 30px;
+    background-color: transparent;
+    padding: 0;
+    border: 0;
+  }
+  .v-select .dropdown-toggle .clear .cross{
     font-size: 23px;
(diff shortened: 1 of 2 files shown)


src/components/Select.vue
```diff
@@ -345,16 +349,17 @@
...
-        type="button" 
-        class="clear" 
-        title="Clear selection" 
+        type="button"
+        class="clear"
+        title="Clear selection"
       >
-        <span aria-hidden="true">&times;</span>
+        <slot name="clear">
+          <span class="cross" aria-hidden="true">&times;</span>
```
stevenharman: Is the `class=“cross”` is needed here? Or perhaps a more explicit name for styling would be better?

alvarosaburido: maybe it will more scalable to have a slot for the `vs__actions` itself

PR #560, opened 2018-06-08
Slots for list header and footer

jacobg:


src/components/Select.vue
```diff
@@ -366,6 +366,7 @@
 
     <transition :name="transition">
       <ul ref="dropdownMenu" v-if="dropdownOpen" class="dropdown-menu" :style="{ 'max-height': maxHeight }">
+        <slot name="listHeader"></slot>
```
stevenharman: For consistency the slots should probably be an `<li>` and have a class denoting headed/footer. Yeah?

sagalbot: @jacobg interested, just backlogged!

PR #589, opened 2018-07-16
implemented disable item

vitaliypva:
added property itemDisabled, with default value isDisabled


Zero3009:
Please resolve this push

PR #594, opened 2018-07-18
Use a Flexbox-based Layout

stevenharman:
This change moves away from floats and absolute positioning in favor of flexbox. Flexbox allows us to solve some of the more quirky issues we're having with elements (e.g, the input) being too big, causing "extra line breaks", vertical alignment of close buttons, etc... and simplified RTL support!

I did need to introduce two new child elements to the `dropdown-toggle` element. These are used to group all of the selected tags and the input in one group. And the "actions" (clear button, dropdown indicator, and spinner) in another. Doing so has the added benefit of no longer allowing selected tags from running "under" those “action” elements.

**NOTE:** The large blocks of change are due to white space differences from indenting inside those new wrapper elements. [View the diff ignoring white space](https://github.com/sagalbot/vue-select/pull/594/files?w=1) to see a more accurate representation of the change here.

This change should apply fairly cleanly atop #512. But I'm happy to rebase this atop that one once it's merged.

**BEFORE**

vue_select_dev

**AFTER**

vue_select_dev

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -224,17 +224,12 @@
     cursor: pointer;
   }
   .v-select input[type="search"].hidden {
-    width: 0px;
+    height: 0;
     padding: 0;
-  }
-  .v-select input[type="search"].shrunk {
-    width: auto;
-  }
-  .v-select input[type="search"].empty {
-    width: 100%;
+    width: 0;
   }
 
-    /* List Items */
+  /* List Items */
   .v-select li {
     line-height: 1.42857143; /* Normalize line height */
   }
@@ -1002,9 +997,7 @@
        */
       inputClasses() {
         return {
-          hidden: !this.multiple && !this.isValueEmpty && !this.dropdownOpen,
-          shrunk: this.multiple && !this.isValueEmpty,
-          empty: this.isValueEmpty,
+          hidden: !this.isValueEmpty && !this.dropdownOpen
         }
       },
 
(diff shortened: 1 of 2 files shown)


src/components/Select.vue
```diff
@@ -967,7 +978,7 @@
           searchable: this.searchable,
           unsearchable: !this.searchable,
           loading: this.mutableLoading,
-          rtl: this.dir === 'rtl',
+          rtl: this.dir === 'rtl', // This can be removed - styling is handled by `dir="rtl"` attribute
```
stevenharman: I'm OK to remove this now, since we no longer depend on it internally. But that could be a breaking change if someone did depend on it. Thoughts?

stevenharman:
Let's get #512 merged first. And then I'll rebase this atop it, and make a few more tweaks to account for "hiding" the input and such that's happening with that PR.

src/components/Select.vue
stevenharman: @sagalbot Should we keep this class around for backward compat, or kill it?

stevenharman:
@sagalbot I _think_ I've covered all of the adjustments needed to account for #512. I'm going to pull this into my app and look for oddities, and push any tweaks that come out of that. In the mean time, would you like to start having a look/playing with these changes and sending along feedback?

eriknygren:
Nice work @stevenharman, this is a lot more versatile as a solution and can cater for the spacing issues in multiple mode better. Definitely the way to move forward 👍

stevenharman:
@sagalbot I've added you as a contributor to that repo.

stevenharman:
@sagalbot Any ideas when we'll see this go out in the `v2.5.0` release? It would be nice to not have to point at a custom branch...

PR #595, opened 2018-07-18
Clicking a Selected Tag Toggles the Dropdown

stevenharman:
I'm not sure how you feel about looking specifically for the `selected-tag` class when filtering the `toggleDropdown` event handler, but targeting it specifically seemed in line with the other handlers. But because the tags are repeated, and not directly part of a `v-for`, we can't use `$refs` to access them. This seemed an OK compromise... 🤷‍♀️

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -847,7 +847,8 @@
        * @return {void}
        */
       toggleDropdown(e) {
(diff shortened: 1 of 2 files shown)


sreenadh:
@sagalbot please release 2.5.2 with this fix

sreenadh:
Got a workaround
.v-select .selected-tag {
    position: absolute;
    pointer-events: none;
    top: 8px;
}

PR #599, opened 2018-07-27
Disable autocomplete with updated chrome method

joemsak:
Chrome ignores `autocomplete=(false | off | random-string)`

In order to leave you be, their latest docs seem to suggest that if you **name** the field as `password` and then set autocomplete to `new-password` it will not attempt to run autocomplete on the field

This has been tested in the latest version of chrome on OS X by me (but not other browers yet, I would be happy to if asked)

I am not sure if there is another combination that would work besides naming the field password, as I would hate to cause bad implications from using this name on the field.

Any suggestions are welcome


eriknygren:
@joemsak Have a look at my proposed solution please: https://github.com/sagalbot/vue-select/pull/730

PR #607, opened 2018-08-01
Added combobox, listbox, and option roles, Closes #606

### LINKED ISSUE #606: Add aria roles to input, list, and options.
Input, list, and list items should have roles of combobox, listbox, and option respectively per W3C aria documentation.

https://www.w3.org/WAI/PF/aria/roles#combobox

```html
<input type="text" aria-label="Tag" role="combobox" aria-expanded="true"
  aria-autocomplete="list" aria-owns="owned_listbox" aria-activedescendant="selected_option">
<ul role="listbox" id="owned_listbox">
  <li role="option">Zebra</li>
  <li role="option" id="selected_option">Zoom</li>
</ul>
```

Will be issuing pull request along with issue

beattyml1:
Added combobox, listbox, and option roles, to input, ul, and li elements respectively per W3C spec, Closes #606

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -355,6 +355,7 @@
               :tabindex="tabindex"
               :readonly="!searchable"
               :id="inputId"
+              role="combobox"
               aria-label="Search for option"
       >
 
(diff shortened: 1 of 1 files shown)


beattyml1:
One thing I realized after looking at the spec again after issuing the pull request is that the aria owns attribute is considered required. It shouldn't hurt anything that it's not there but I'm still planning to look at a fix since I'm not sure screen readers will really get the full benefit until it's taken care of

beattyml1:
Preference for generating unique IDs between:
Random 9 character alphanumeric
```js
Math.random().toString(36).substr(2, 9)
```
VS

Incremented Integer
``` JS
let id = 0;
...
export defualt {
  ...
  data() {
    return {
      ...
      uniqueId: id++
    }
  }
}
```

I should be able to get the whichever is prefer done super quick

PR #609, opened 2018-08-05
Added aria-owns link between input and dropdown list

beattyml1:
Added aria-owns link between input and dropdown list, had to add unique ID generator and a switch v-if to v-show to make that really work.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -357,6 +357,7 @@
               :id="inputId"
               role="combobox"
               :aria-expanded="dropdownOpen"
+              :aria-owns="listUniqueId"
               aria-label="Search for option"
       >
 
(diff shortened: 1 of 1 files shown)


beattyml1:
I assumed it was but as I think about it I feel like if it's not there while the screen reader may not feel existentially happy it also probably isn't going to cause any bugs. My guess is technically invalid but all screen readers would still do their thing which to me is what we really care about. I'll switch it back to v-if and recheck the screen reader.

beattyml1:
A quick update I've determined that v-shows definitely doesn't work quite as well as v-if and have tried to find a work around. I also tried doing some further screen reader testing and covered another screen reader issue that I was planning to resolve at the same time and update the PR when I get a chance.

PR #611, opened 2018-08-10
Bug/hidden input

sagalbot:
Couldn't click on input after a value was selected due to a 0 height. Removed the css rule.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -216,7 +216,6 @@
   }
   .v-select input[type="search"].hidden {
     border: none;
-    height: 0;
     padding: 0;
     width: 0;
   }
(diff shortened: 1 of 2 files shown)


src/components/Select.vue
```diff
@@ -216,7 +216,6 @@
   }
   .v-select input[type="search"].hidden {
     border: none;
-    height: 0;
```
stevenharman: I seem to recall adding this for a reason... but now I can't recall what it was. In any case, I don't understand why you'd not be able to click to trigger the dropdown. The actual `input` is already `width: 0` in the case of it being hidden. Hmm....

stevenharman:
I was thinking the clock handler was at a higher level, and thus we didn’t actually need the input to be of any significant size.

PR #615, opened 2018-08-13
WIP: Update Build to Vue CLI 3.0

sagalbot:
- updates previous webpack build system to use `vue-cli-service build`
- updates dev environment, adds build command for netlify PR previews
- adds linting
- migrates test suite onto Jest / vue-test-utils

---

### ToDo

- [ ]  `/dist` output has changed, verify effects
- [ ] migrate docs/homepage to VuePress
- [ ] (optional) update dev environment to show only a single select, but all options are configurable
- [ ] add tests for #703


luishdz1010:
This refactoring seems to make possible including `vue-select/src/index.js` again for Nuxt2/Webpack 4, which was broken due to default exports using `module.export`. Thanks for the hard work!

yamsellem:
IMHO this PR is very ambitious: migrating from webpack to vue-cli 3, from karma to jest, from gitbook to vuepress.

@sagalbot may you consider splitting it? It has great value, but is somehow blocked due to is ambition.
I've just added a PR with only webpack 4 (related to SSR), maybe it's possible to release jest tests & lint first. Then merging webpack 4 or vue-cli 3 (testing the bundle with SSR and modules imports, in both cases I guess).

The vuepress may be handled later, I suppose.

It really seems to me that this migration is almost done, and it is maybe preventing some more to come. I'll be happy to here you on the matter.

Take care.

PR #626, opened 2018-09-02
Added on-input to update search to fix mobile

MarkSG93:
This fixes the typeahead results not updating on mobile. Related to https://github.com/vuejs/vue/issues/8231 which is a bug in VueJS.


MarkSG93:
Hey @sagalbot, sorry didn't get an email note you replied. Yes it should do! I've deployed an Vue app to production using my PR and I've seen no regressions and it now works correctly on mobile.

PR #628, opened 2018-09-05
Add option group feature

Ovski4:
Hi,

I have been working on a project using this vue-select component. I now need to group the options but this feature is not currently supported.

I would appreciate if you could give me your thoughts on this PR. It's doing the job for me without breaking changes.

Thank you for your time!


smholsen:
Any reason why this was closed? Any improvement suggestions? Would be really great to get this in

PR #631, opened 2018-09-06
Bug/629 show selected value

### LINKED ISSUE #629: Selected value no longer shows when opening dropdown
With the release of 2.5, if I select a value, and then re-open the dropdown, the selected value no longer stays showing in the input box:

[Codepen](https://codepen.io/anon/pen/eLEgor)

2 5-value-disappears

Previously, the value would stay showing, but be grayed out:

[Codepen](https://codepen.io/anon/pen/dqzvKM)

2 4-value-grays-out

I'm not sure if this was an intentional change, it looks to have stemmed from the [change to flexbox](https://github.com/sagalbot/vue-select/pull/594), but I don't see any discussion around this feature specifically.

I was looking at switching to vue-select from vue-multiselect specifically because vue-multiselect does not have a way to keep the selected value visible when opening the dropdown, so it would be great to have this feature available again in vue-select.

sagalbot:
Ensures that the input displays the currently selected value of single selects while dropdown is open. @stevenharman this PR removes the `hidden` class. I did a bunch of testing, it looks like we just don't need it. Any red flags?

- removes `hidden` class & `inputClasses` property
  - removes associated tests
- fixes minor alignment issue

## Before

before

## After

after

---

Closes #629 
Related #621

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -196,20 +196,13 @@
     font-size: 1em;
     display: inline-block;
     border: 1px solid transparent;
+    border-left: none;
     outline: none;
     margin: 4px 0 0 0;
-    padding: 0 0.5em;
+    padding: 0 7px;
     max-width: 100%;
     background: none;
     box-shadow: none;
-
-    /* `flex-grow` will stretch the input to take all remaining space, but We
-       need to ensure a small amount of space so there's room to type input. We'll
-       set the input to "hidden" (via width: 0) when the dropdown is closed, to
-       prevent adding a "blank" line (see: https://github.com/sagalbot/vue-select/pull/512).
-       In that case, the flex-grow will still stretch the input to take any
-       available space, on the same "line."
-    */
     flex-grow: 1;
     width: 4em;
   }
@@ -218,11 +211,6 @@
   }
   .v-select.unsearchable input[type="search"]:hover {
     cursor: pointer;
-  }
-  .v-select input[type="search"].hidden {
-    border: none;
-    padding: 0;
-    width: 0;
   }
 
   /* List Items */
@@ -348,7 +336,6 @@
                 @focus="onSearchFocus"
                 type="search"
                 class="form-control"
-                :class="inputClasses"
                 autocomplete="off"
                 :disabled="disabled"
                 :placeholder="searchPlaceholder"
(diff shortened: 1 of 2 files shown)


stevenharman:
How do multi-selects react to the change? It might all be a non-issue now that we have the new layout mechanism, but at one point the `.hidden` helped prevent an input from going to multiple lines when there was still "space" on the first line.

I'm going to pull the code and run the dev page locally, to play around a bit too.

src/components/Select.vue
```diff
@@ -191,20 +196,13 @@
...
     background: none;
     box-shadow: none;
-
-    /* `flex-grow` will stretch the input to take all remaining space, but We
-       need to ensure a small amount of space so there's room to type input. We'll
-       set the input to "hidden" (via width: 0) when the dropdown is closed, to
-       prevent adding a "blank" line (see: https://github.com/sagalbot/vue-select/pull/512).
-       In that case, the flex-grow will still stretch the input to take any
-       available space, on the same "line."
-    */
```
stevenharman: @sagalbot I think this, or something like it, is still needed, or we'll get the extra "blank line".

vue_select_dev

It seems that adjusting the `width: 0;` (two lines down) will fix that problem. In the image below I've done that and outlined the inputs in red, so we can see where they are. Seems to work in all cases?

vue_select_dev

sagalbot: Yeah, I got the same result:

```css
  .v-select input[type="search"],
  .v-select input[type="search"]:focus {
    appearance: none;
    -webkit-appearance: none;
    -moz-appearance: none;
    line-height: 1.42857143;
    font-size: 1em;
    display: inline-block;
    border: 1px solid transparent;
    border-left: none;
    outline: none;
    margin: 4px 0 0 0;
    padding: 0 7px;
    max-width: 100%;
    background: none;
    box-shadow: none;
    flex-grow: 1;
    width: 0;  /* 0 here instead of 4em */
  }
```

### Result

image

I don't understand it, but I'll take it! 🤷‍♂️

stevenharman: I *think* it's overriding the browser's default behavior of growing the input to some minimum width. I've tried setting `min-width: 0` and `flex-basis: 0` instead, but they didn't work. So...as you say, 🤷‍♀️.

stevenharman:
@sagalbot can we get this fix out in a `2.5.1` release?

PR #633, opened 2018-09-06
Add value property

baranovxyz:
Currently it is not possible to get just values from options objects.
This PR fixes it by adding prop 'valueProp' which access options objects properties.


ddzobov:
I think that add it to docs extremely important 😊

agonzalez0515:
Would really appreciate this documentation! I don't see the value prop being used in that codepen.

dbarrington:
I expected value to work in this case too, but adding index prop worked for my object. 

```
<v-select maxHeight="200px" v-model="form.target_id" index="id" label='name' :options="targets" >
</v-select>
```

geri777:
The index property worked for me:

`<v-select v-model="vsel" index="value" label="label" :options="[{value:0, label:'Hello']">`

PR #658, opened 2018-09-24
Support for optgroups

azriel46d:
I've added support for optgroups 
Optgroups can be passed as part of the options with the following structure
```
{
   title: "lorem ipsum" // This is the title of the opt group
   option: []  // the options in that opt group, either as strings or as objects
}
These can be mixed with items not in any opt groups.

What i did is that before setting the mutableOptions, the optgroups are flattened out using the function ```normaliseOptGroups``` . 
In the template there is a check to see if the item is an optgroup to just render a non clickable list item.

### DIFF
--- a/dev/dev.html
+++ b/dev/dev.html
@@ -31,6 +31,7 @@
 <body>
     <div id="app">
         <v-select placeholder="default" :options="options"></v-select>
+        <v-select placeholder="optgroups" :options="optgroups"></v-select>
         <v-select placeholder="default, RTL" :options="options" dir="rtl"></v-select>
         <v-select placeholder="default, options=[1,5,10]" :options="[1,5,10]"></v-select>
         <v-select placeholder="multiple" multiple :options="options"></v-select>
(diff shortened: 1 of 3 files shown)


jasontxf:
Hello, Any update to this? I came upon a use case where I need optgroup too.

bfic:
@sagalbot I am the next person who needs optgroup too

jd-0001:
+1

CeloHill:
+1 pleeease, we all need optgroups 
and thanks for the component

leoclo:
+1

gr8bit:
@sagalbot Hi Jeff, do you have any idea on how you would like to have option groups implemented into vue-select? I'd be happy to code it the way you think fits best into vue-select. We're currently updating our own branch with this PR but I'd rather like to see this (HTML select) feature become part of the project. :) Any input appreciated!

howard-tzw:
Now you can make use of `optgroups` in vue3-select@0.1.0, which is a forked repository exclusively maintained for Vue 3

- repo: https://github.com/howard-tzw/vue3-select
- demo: https://vue3-select.vercel.app/optgroups

PR #669, opened 2018-10-04
Fix bug with filterable prop on taggable inputs

andywarren86:


src/components/Select.vue
```diff
@@ -612,8 +612,7 @@
 
       /**
        * When true, existing options will be filtered
-       * by the search text. Should not be used in conjunction
-       * with taggable.
+       * by the search text.
```
andywarren86: I couldn't see any reason why taggable & filterable can't be used together anymore. Let me know if I'm wrong.

src/components/Select.vue
```diff
@@ -1112,10 +1111,14 @@
        * @return {array}
        */
       filteredOptions() {
-        if (!this.filterable && !this.taggable) {
-          return this.mutableOptions.slice()
+        let options = [...this.mutableOptions];
```
andywarren86: This makes a shallow copy which I assume is what the `this.mutableOptions.slice()` was doing before.

test/unit/specs/Select.spec.js
```diff
@@ -1295,12 +1295,11 @@ describe('Select.vue', () => {
 			})
 		})
 
-		it('should select an existing option if the search string matches an objects label from options when filter-options is false', (done) => {
-			let two = {label: 'two'}
+		it('should select first option if the search string matches an objects label from options when filterable is false', (done) => {
```
andywarren86: Not sure what to do with this one. Previously the filter would be applied leaving only the matched option at the top of the list. Now the filter is not longer being applied the option list remains unchanged so the first option would be selected.

The only other way I can think to handle this would be the re-order the options such that the matched option is moved to the top of the filteredOptions list?

Either way it is bit of an edge case -- with `filterable=false` you would expect the server to provide already filtered options.

PR #680, opened 2018-10-14
Disable individual options

anzorasadov:
Added a possibility to disable individual options rather than the whole Select element. 

Usage: 
```js
<v-select
  ref="select"
  :options="[{label: 'foo', value: 'foo', disabled: true},{label: 'bar', value: 'bar'}]"
></v-select>
```
in this example the option `foo` will be disabled


anzorasadov:
@sagalbot I am not sure what is wrong with the deployment script. But it seems like all the recent PRs are experiencing the same issue.

jtpdev:
@anzorasadov I have the same problem with my pull request. And I need your funcionality too.

gr8bit:
Please do not let this very useful feature hang here @sagalbot <3

vargasroger:
Hello there, this is a cool feature. We kind need this here.
It seems this PR is forgotten.

@sagalbot  If is there a way to help let me know.

PR #683, opened 2018-10-15
Create index.d.ts

Pasadenka:
Base typescript declaration for typescript error fix.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1 @@
+declare module 'vue-select';


zlumer:
TypeScript typings are available on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vue-select/index.d.ts).
Install them via `yarn add @types/vue-select`
Although I had to cast `VueSelect` to `any` before registering the component, so there's definitely room for improvement.

andydlindsay:
@Pasadenka @zlumer This appears to only be an issue when installing with `yarn`. If you install with `npm`, then you don't have to cast to `VueSelect` to `any`.

I'm afraid that I don't know enough about the differences between `yarn` and `npm` to understand why this happens.

viT-1:
Any news about d.ts (@types/vue-select) on v3?

viT-1:
@sagalbot  I'm new in ts too =)
But as a user of your package imported in ts files I create vue-select/index.d.ts file in typings folder wihich is configured in tsconfig.json typeRoots section as ```
"typeRoots": [
			"../node_modules/@types",
			"./typings"
		]
```

index.d.ts has only one row of code with declaration for all as "any"
`declare module 'vue-select';`
but should more [e.g. for v2.3](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vue-select/index.d.ts)

Issue #689, opened 2018-10-22
Component stuck after click "Sorry, there are no match options" option


italoSilveira:
Hey there,
I was playing with the component and saw this bug. When I click the "Sorry, there are no match options"option, the component hangs:
![vue-select-issue](https://user-images.githubusercontent.com/32437227/47299545-d2c82700-d5f0-11e8-9dc4-b0cab2c308dc.gif)

You can see the code here and try ourself: https://codepen.io/sagalbot/pen/XVoWxm

haniefutama:
I have this issue as well. No console out put as well. Anybody has an idea?

italoSilveira:
Meanwhile, someone has found a temporary solution https://github.com/sagalbot/vue-select/issues/648#issuecomment-424812677 :

```
<template slot="no-options">
        <div @mousedown.stop="">
          Begin typing to search (at least 3 letters)
        </div>
</template>
```

Issue #648, opened 2018-09-17
the components can't focus  and the ropdown can't  close when clicking 'Sorry, no matching options'


hellolex:
Is there any solution?

![bandicam 2018-09-17 16-48-42-908](https://user-images.githubusercontent.com/22807285/45614868-775cc380-ba9d-11e8-93de-bec8b4b62e8e.gif)

Pyro979:
This is actually reproducible on one of the examples - [Vue Select - Ajax](https://codepen.io/sagalbot/pen/POMeOX)

Pyro979:
Proposed PR: https://github.com/sagalbot/vue-select/pull/664

Meanwhile, if someone is having this issue in production - wrap the content of the slot with a div and prevent the click there.

```
<template slot="no-options">
        <div @mousedown.stop="">
          Begin typing to search (at least 3 letters)
        </div>
      </template>
```

DreaMinder:
this happens not only with 'no-options'. If you click on scrollbar, dropdown will stuck too.

sagalbot:
Fixed by #703

Pyro979:
@sagalbot 👍 Do you know when this is going to make it into a release?

sagalbot:
@Pyro979 hoping to push out 3.0 by the end of the week. It's a major bump, but the breaking changes should be pretty minimal. So far the only breaking change will be that CSS is separate from the JS build.

PR #691, opened 2018-10-23
Fix inability to interact with vue-select after clicking the "No-options" placeholder

Almoullim:
Fix inability to interact with vue-select after clicking the "No-options" placeholder
Issue: [#689](https://github.com/sagalbot/vue-select/issues/689#issue-372556564) and [#648](https://github.com/sagalbot/vue-select/issues/648#issue-360784044)

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -1112,7 +1112,7 @@
        * @return {array}
        */
       filteredOptions() {
-        if (!this.filterable) {
+        if (!this.filterable && !this.taggable) {
           return this.mutableOptions.slice()
         }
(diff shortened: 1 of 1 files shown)


eriknygren:
Cool! Had a test, fixes the "freeze" issue. I don't understand why the `&& this.searching` or the opposite is part of that conditional at all though, do you know?

MrStobbart:
I created a similar pull request (#703) that removed the check on `this.searching` completely as the here suggested solution did not work if a vue-select had no options to begin with but the user already typed some text.

@eriknygren I also don't know why there was a check for `this.searching` here as I think a user would want the element be unfocussed in any case (`this.$emit('search:blur')`) and not only when he did search and had no mousedown.

PR #693, opened 2018-10-25
Fixes #671. The method isOptionSelected ...

### LINKED ISSUE #671: multiple options (objects) does not disable selected options properly
When I am using multiple options (objects) it does not disable the selected options properly. It is possible to select the same option multiple times.

I have created a demo on codepen: https://codepen.io/telcy/pen/KGMqgm

totola-clx:
Fixes #671. The method isOptionSelected on options of type 'object' was cycling but no returning after an existing option was found (true), resetting to false on next option.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -904,18 +904,12 @@
        * @return {Boolean}        True when selected | False otherwise
        */
       isOptionSelected(option) {
-        let selected = false
-        let i = 0
-        while (!selected && i < this.valueAsArray.length) {
-          let value = this.valueAsArray[i]
+        return this.valueAsArray.some(value => {
           if (typeof value === 'object') {
-            selected = this.optionObjectComparator(value, option)
-          } else if (value === option || value === option[i]) {
-            selected = true
+            return this.optionObjectComparator(value, option)
           }
-          i++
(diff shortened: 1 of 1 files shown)


Magnum5234:
you can use some
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

```javascript
isOptionSelected(option) {
  return this.valueAsArray.some(value => {
    if (typeof value === 'object') {
      return this.optionObjectComparator(value, option)
    }
    return value === option || value === option[this.index]
  })
},
```

totola-clx:
Thank you @Magnum5234 , I didn't know the "some" method.
I updated the pull request with your tip.

totola-clx:
Unfortunately the build fails for some reason, but I cannot understand why... :-(

totola-clx:
Hi @sagalbot ,
please can you add more details on how should I update the PR?
Can you reference some guide or wiki page with instructions on how to add test coverage?
Moreover, the automatic build process has failed, but the reason is not clear to me.
The error reported "5:30:31 PM: error Command "build:dev" not found." seems not related to code.
Thanks in advance,
Cristian

PR #694, opened 2018-10-25
#671 Bug fix

ddzobov:
Fix bug when options are objects and already selected options is selectable


Magnum5234:
you might want to look at https://github.com/sagalbot/vue-select/pull/693/

PR #700, opened 2018-11-08
fix: mousedown to click + without prevent.stop

cybermerlin:
mod: use .set(), .isEmpty on value if exist


src/components/Select.vue
```diff
@@ -369,12 +406,15 @@
...
-          <a @mousedown.prevent.stop="select(option)">
-          <slot name="option" v-bind="(typeof option === 'object')?option:{[label]: option}">
-            {{ getOptionLabel(option) }}
-          </slot>
+      <ul ref="dropdownMenu" v-if="dropdownOpen" class="dropdown-menu" :style="{ 'max-height': maxHeight }"
+          role="listbox" @mousedown="onMousedown">
+        <li role="option" v-for="(option, index) in filteredOptions" :key="index"
+            :class="{ active: isOptionSelected(option), highlight: index === typeAheadPointer }"
+            @mouseover="typeAheadPointer = index">
+          <a @click="select(option)">
```
PieterWigboldus: Fixed also in #814

PR #701, opened 2018-11-09
Tagging UX improvement: Refocus on input after adding a tag in noDrop

eriknygren:
What
---
- If the `noDrop` prop is set to true, and the user is typing in tags,
always refocus the input after each added element.

Why
---
Because at the moment the user needs to click on the element after
adding each tag, this allows for a more fluid user experience.

### Before:
vue select before

### After:
vue select after

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -877,6 +877,9 @@
         if (this.clearSearchOnSelect) {
           this.search = ''
         }
+        if (this.noDrop) {
+          this.focusInput()
+        }
       },
 
       /**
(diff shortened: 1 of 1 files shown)


mylescc:
👏

eriknygren:
I'm gonna close this. In retrospective, I don't feel super confident it's not gonna introduce some other issue and it's a pretty niche improvement.

eriknygren:
@sagalbot 👍 It was so long ago I opened this and I had some kind of vague memory of this causing something unexpected. I just did some retesting and can't seem to figure out what that was haha.

I've amended the PR slightly, to further lock it down explicitly to this use case by only refocusing if `this.noDrop && this.multiple`. So should be all good now.

Couldn't reopen this PR because I rebased the branch, so opened a new one https://github.com/sagalbot/vue-select/pull/755/files

Issue #529, opened 2018-05-02
Multiple select and Vuex


Takitaf:
Got one problem, when storing value of multiple select in store. 

Let's say we have store like this: 
```javascript
state: {
    values: null,
    options: ['a', 'b', 'c']
},
mutations: {
    setValue (state, values) {
        state.values = values;
    }
}
```
and somewhere in the form.vue file:
```javascript
<template>
   <vue-select 
        multiple :value="$store.state.values" 
        :options="$store.state.options"
        @input="$store.state.commit('setValue', $event)"
   >
</template>
```
Of course that's just a simplified example. In `Select.vue` component you have a `select` method, where you have this code:
```javascript
if (this.multiple && !this.mutableValue) {
    this.mutableValue = [option] //FIRST_OPTION_PICK
} else if (this.multiple) {
    this.mutableValue.push(option) //SECOND_OPTION_PICK
} else {
    this.mutableValue = option
}
```

At first change, when we choose a value, everything works fine. That's because FIRST_OPTION_PICK line is executed, new array is created, nothing is mutated. However, when we pick the second value, the SECOND_OPTION_PICK is executed, which just push new value to existing array. And this operation triggers Vuex error `Error: [vuex] Do not mutate vuex store state outside mutation handlers.`

The problem is probably in created method, where at the beginning we have this: `this.mutableValue = this.value`, which creates only a reference, not a copy, when the value is an array. Then in SECOND_OPTION_PICK there is array push, which mutates value array in store.

 Is there any way to make "values" immutable, or it is bug that needs to be fixed in vue-select package?

kotarCreative:
You can handle this by moving the options into a computed value. That way you can define a getter and setter and do not need to include the input event. It could look something like this.
```javascript
computed: {
    options: {
        get() {
            return this.$store.state.options;
        },
        set(options) {
            $store.state.commit('setValue', options);
        }
    }
}```

Then the component can be defined like so `<vue-select :options="options" multiple></vue-select>`

CaleCrawford:
I am having this problem as well. It appears to only be an issue when you have `strict: true` for vuex with `multiple` on the v-select. Here is a reproduction link [https://codepen.io/CaleCrawford/pen/yEYbRP](https://codepen.io/CaleCrawford/pen/yEYbRP). If you add one all is good, if you add two you get the `Error: [vuex] Do not mutate vuex store state outside mutation handlers.` Seems like it coming from inside the vue select lib.

[...]

PR #702, opened 2018-11-09
multiple mode - add/remove actions to support vuex strictmode

eriknygren:
What
---
- In multiple mode: Update adding or removing a value to use
spread operator or filter, instead of mutating the array directly.
This makes it play nicely with `:value="value" @input="onChange"` pattern

Why
---
So vuex strict mode doesn't throw warnings.

https://github.com/sagalbot/vue-select/issues/529

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -826,7 +826,7 @@
           if (this.multiple && !this.mutableValue) {
             this.mutableValue = [option]
           } else if (this.multiple) {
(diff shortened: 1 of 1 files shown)


renaudaste:
@eriknygren thank you for this fix :) 
Any news about a possible merge ?

eriknygren:
@renaudaste no worries! No news I'm afraid for now. It depends on whenever @sagalbot is around to review/merge.

PR #703, opened 2018-11-15
Fixed unresponsive vue-select after clicking custom no-options (#689, #648, #722, and #736) as well as isOptionSelected wrong check for object (#732)

MrStobbart:
Fixes the the bugs #689 and #648 by making the vue-select still interactive after clicking a custom no-options message. To test this another vue-select was inserted into the test dev.html

This merge request is similar to #691 but prevents the bug when the users has typed something but the no-options message still appears.


eriknygren:
Nice one @MrStobbart. I had a trawl through the git log to see the original reason for that if statement: https://github.com/sagalbot/vue-select/commit/350998b2da75773a69843143eab897af4f6efb16

Is this something that might have been reintroduced as a side effect?

MrStobbart:
@eriknygren Uh thanks for looking that up (should have thought about that myself).
I will do some tests tomorrow and check if that bug was reintroduced. If so I will try to find a solution for that. :)

MrStobbart:
@eriknygren My first fix reintroduces the bug. I found a different solution, that is now also part of this pull request, that should close the issues #689 and #648 and does not reintroduce the IE bug that the commit https://github.com/sagalbot/vue-select/commit/350998b2da75773a69843143eab897af4f6efb16 fixed.

I tested the solution in Internet Explorer 11, Safari, Chrome, Firefox, Opera and Edge. 
The behaviour when just clicking the vue-select without having typed something beforehand is still a bit odd, as the options can only be closed by selecting one of them. A solution for that would be to only have this behaviour for IE (so that scrolling is possible) and not for the other browsers. What do you think about that?

dev/dev.html
```diff
@@ -75,9 +75,12 @@
         <v-select placeholder="filtering with fuse.js" label="title" :options="fuseSearchOptions" :filter="fuseSearch">
             <template slot="option" scope="option">
                 <strong>{{ option.title }}</strong><br>
-                <em>{{ `${option.author.firstName} ${option.author.lastName}` }}</em>
+                <em>{{ option.author.firstName + ' ' + option.author.lastName }}</em>
```
eriknygren: is there a reason to not use interpolation here? The prior code style was probably the way to go

src/components/Select.vue
```diff
@@ -969,15 +969,31 @@
...
           if (this.clearSearchOnBlur) {
             this.search = ''
           }
-          this.open = false
-          this.$emit('search:blur')
+          this.closeSearchOptions()
+        }
+
+        // Fixed bug where no-options message could not be closed
+        if(this.search.length === 0 && this.options.length === 0){
```
eriknygren: so when a search query doesn't exist, and there are no options available, close the search? Is this maybe an `else if` to be joined up with the above conditional, I'm thinking otherwise `closeSearchOptions` could be called twice.

src/components/Select.vue
```diff
@@ -969,15 +969,31 @@
        * @return {void}
        */
       onSearchBlur() {
+
```
eriknygren: unnecessary whitespace 👾

dev/dev.html
MrStobbart: This file was was never compiled and could therefore not be run in internet explorer. I would leave it like this to make it testable in IE.

src/components/Select.vue
MrStobbart: Yeah good point. Maybe return void after the first time closeSearchOptions has been called? Seems more readable for me then.

src/components/Select.vue
MrStobbart: I will remove that

MrStobbart:
@eriknygren Ok, I would leave it as it is then and this slightly different behaviour could be done in a different pull request when necessary. Thanks for the feedback on the code and your quick responses!

eriknygren:
> @eriknygren Ok, I would leave it as it is then and this slightly different behaviour could be done in a different pull request when necessary. Thanks for the feedback on the code and your quick responses!

No worries, thank you!

MrStobbart:
This pull request also fixes #722, and #736 and I added the suggested fix for #732.

victorsilent:
When it will be released? I have this bug with isOptionSelected

sergeyBreslavets:
try 
 <v-select v-if="drawComponent"      .....
.......
<span slot="no-options" class="bag" @click="reset">
              {{$t('text_start_search')}}
</span>
.......
reset: function() {
        var vm = this;
        console.log('wres')
        vm.drawComponent = false;

        this.$nextTick(function() { .  // for nuxt     
          vm.drawComponent = true;
        });
        
     Vue.nextTick(function () { .   //vue      
    vm.drawComponent = true;
      })     

      },

PR #711, opened 2018-12-10
Only emit input on actual input events

### LINKED ISSUE #455: Selected values get reset right away
So myeditor is the Vue instance.  Inside i have a reusable input-container component that is meant for form inputs. The slot gets the form input and surrounds it with some stuff.
Inside I have a vue-select that has an issue. Whenever I select a value, the value gets displayed for a second or so and then disappears again. My guess is that there is an issue with associating the v-model so that the referenced model does not update. Any ideas?
Ah and 'data' is on 'myeditor'.

`<myeditor class="eventEditor" inline-template>  
    <input-container label="{{'Description'}}" :errors="editor.errors">  
        <template slot="input" slot-scope="slotProps">  
             <v-select ref='locationsSelect' multiple :filterable="false" :options="locationOptions"
                                      v-model="data.locations" @search="onLocationSearch">
        </template>  
    </input-container>  
</myeditor>`

### LINKED ISSUE #504: Input event emits as if it were onchange
In keeping with other vue components and practice - the input event should only be emitted when the user gives the select input with their peripherals, and should not emit when being programmatically changed. This is consistent behaviour.  

Currently it acts more like an OnChange event.

### LINKED ISSUE #526: Input event emits as if it were onchange
Can't seem to re-open #504 - Close if inappropriate, but this behaviour is inconsistent with all other inputs - As mentioned in #504 this change would not break synchronisation. 

You can see this difference by performing a simple test:
```
computed: {
        Ouput: {
            get() {
                console.log("input GET");
                return "1";
            },
            set(v) {
                console.log("input SET");
            }
        },
        Ouput_Select: {
            get() {
                console.log("select GET");
                return "1";
            },
            set(v) {
                console.log("select SET");
            }
        }
    }
```

With an DOM setup like
```
<v-select :options = "['1','2','3','4']" v-model = "Output_Select" />
<input type = "text" v-model = "Output" />
```

When this DOM is rendered the first time you will get in the console:
> input GET
> select SET

Hope this is enough to showcase the inconsistency - and also prove that it won't desync the state!

### LINKED ISSUE #545: @change not working
"vue": "^2.5.16",
"vue-select": "^2.4.0",

```
<script lang='coffee'>
	Vue = require 'vue/dist/vue.js'
	setParams = Vue.component 'set-params',
		props: 
			params:
				type: Array
				required: true

		methods:
			getOptions: (srv) ->
				console.log('WTF')
				@$emit('get-option', srv)
	module.exports = setParams
</script>
<template>
	<div class='accordion__step step v-step'>
		<div class='step__selects'>
			<div class='v-step__select' v-for='select in params'>
				<v-select v-on:change='getOptions(select.get)' class='v-step__v-select' v-if='select.options.length === 0' label='text' :placeholder='select.placeholder' :disabled='true'></v-select>
				<v-select v-on:change='getOptions(select.get)' class='v-step__v-select' v-else label='text' :placeholder='select.placeholder' :options='select.options'></v-select>
			</div>
		</div>
	</div>
</template>
```
console log and emitter not worked. Also I tried v-on:change @select, @selected.
event starts working after @input. I didnt find any info about event API, only prop function "onChange".
pls, add some info about event API

### LINKED ISSUE #573: @input is called on component load?
Is @input event supposed to be called on page load? If so, is it possible to prevent that?
My code : 
`<v-select label="manufacturer" placeholder="Search for equipment..." :value="this.$store.state.selectedEquipment" :options="this.$store.state.equipment" @input="setSelectedEquipment($event)"></v-select>`

`setSelectedEquipment(e){
				this.$store.dispatch('selectEquipment',e)
				this.$router.push('equipment')
			}`

### LINKED ISSUE #659: How to add default value without triggering @input event?
I want to implement a route change prevention if the changes are made to any values in my form. The problem that i have with v-select is that i don't seem to find a way to check if changes are made correctly.

```
<v-select
        :options="categoryOptions"
        v-model="categoryValue">
</v-select> 
```
When the selected value is populated from the vuex state an @Input event is emitted and then set is called and so my isDirty value is changed to true.

```
computed: {
 categoryValue: {
        get () {
          return this.$store.getters[...]
        },
        set (value) {
          this.isDirty = true
          this.$store.commit(..., value)
        }
      },
},
```

Is there a way to set a default value for vue-select that does not trigger the @input event or any other way to implement this?

TaylorDale:
Separated the input event from the change event, which will still occur upon any change of value (programmatic, user, whathaveyou).

The input event now only occurs upon user input as is standard with all other components and form fields. 

This resolves #659
Resolves #573 
Resolves #545 
Resolves #526 
Resolves #504 
Resolves #455

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -566,7 +566,14 @@
       onChange: {
         type: Function,
         default: function (val) {
-          this.$emit('input', val)
+          this.$emit('change', val);
+        }
+      },
+
+      onInput: {
+        type: Function,
+        default: function (val) {
+          this.$emit('input', val);
         }
       },
 
@@ -830,6 +837,7 @@
           } else {
             this.mutableValue = option
           }
+          this.onInput(this.mutableValue);
         }
 
         this.onAfterSelect(option)
@@ -853,6 +861,7 @@
         } else {
           this.mutableValue = null
         }
+        this.onInput(this.mutableValue);
       },
 
       /**
@@ -861,6 +870,7 @@
        */
       clearSelection() {
         this.mutableValue = this.multiple ? [] : null
+        this.onInput(this.mutableValue)
       },
 
       /**


rurodev:
Would need exactly this change to use this component, what are the chances to have it merged? Otherwise I need to look for an alternate.

shredmaster:
i also need this feature to use this component. anyone have a workaround?

TaylorDale:
You can just install my fork with Npm to use it right now

angeliski:
Hey @sagalbot Any update about that merge? You need some help about that?

PR #725, opened 2019-01-08
don't pre-highlight on focus, because if it's after a search, it's possibly wrong

killroyboy:
This problem is mostly apparent when using `selectOnTab`, because tabbing into and out of the field would change the value it to the highlighted option.


killroyboy:
@sagalbot I'm fine with that distinction. We only run into this problem with selectOnTab is true.

Issue #736, opened 2019-01-23
Vue-Select breaks when "No options available" is clicked


Fusseldieb:
Well, this issue happens when I use `taggable` and click on the "Sorry, no matching options".

As soon as I click that, that particular select freezes and only a page reload fixes it.

It can be seen even on the stock settings here: https://sagalbot.github.io/vue-select/docs/Basics/Values.html#tagging

EDIT: A quick GIF to show the issue easier:
![bugselect](https://user-images.githubusercontent.com/4715129/51620406-73442500-1f19-11e9-971e-317fc6597dce.gif)

Fusseldieb:
Solved it myself by fixing and rebuilding it.

If somebody need this fix ASAP (too), you can do:

npm uninstall vue-select
npm install --save https://github.com/Fusseldieb/vue-select
But please note that this will then link to my version of the vue-select, which I don't guarantee to stay up to date. As long as I use this package, I'll keep it up to date, but I don't guarantee it! Please roll "back" to the official version if mine goes behind or if this pull request gets approved with npm uninstall vue-select && npm install vue-select

sagalbot:
Fixed by #703

Issue #722, opened 2019-01-04
Dropdown doesn't close when there is a large amount of data


arethk:
You can see the problem in your codepen example: https://codepen.io/sagalbot/pen/opMGro

[...]

PR #737, opened 2019-01-24
Solved blur freeze issue

Fusseldieb:
Fixed a issue which prevented the user from clicking away in certain circumstances (& other things)

Essentially solves issues #722 and #736 

**EDIT:** The _netlify_ build won't pass, but I don't think that's a problem, since that particular test passes on nobody here. 

**EDIT2:** I, as a new git user, unfortunately referenced this PR to my master branch, so there are more fixes & features to be expected in _this_ same PR. I know, that's not optimal, sorry.


src/components/Select.vue
```diff
@@ -318,7 +318,7 @@
           :multiple="multiple"
           :disabled="disabled"
         >
-          <span class="selected-tag" v-bind:key="option.index">
+          <span class="selected-tag" v-bind:key="option.index" :class="{ 'multiple':multiple }">
```
Fusseldieb: This particular change allows to better style a CSS for it. In my usage, I use "chips" for each multiple input, but not on regular inputs. To differentiate and style both, this is very useful.

src/components/Select.vue
```diff
@@ -318,7 +318,7 @@
           :multiple="multiple"
           :disabled="disabled"
         >
-          <span class="selected-tag" v-bind:key="option.index" :class="{ 'multiple':multiple }">
```
Fusseldieb: I discovered that the v-select already has a CSS class which differentiates both.
To style the multiple input, I now use:
```
.v-select:not(.single) .selected-tag {
}
```

PR #741, opened 2019-01-31
Prop `texts` to customize vue-select texts (ie. internationalization)

sne3ks:
Hello, 

I use vue-select on a multilanguage application, and i had some difficulties to translate hardcoded texts. For example _"Clear selection"_ is a ```title```  on button or _"Search for options"_ an ```aria-label``` on input. 

I know that for the others texts slots can be used. However we use v-select many and many times, and it become repetitives to write slots for each v-select instance.

With the ```texts``` props, we can write this snippet and translate / customize text once and for all when importing.

```js 
import vSelect from 'vue-select'

vSelect.props.texts = {
    type : Object, 
    default : () => ({
         noResults     : i18n.t("vSelect_noResults"),
         loading       : i18n.t("vSelect_loading"),
         searchOption  : i18n.t("vSelect_searchOption"),
         clearSelection: i18n.t("vSelect_clearSelection"),
    })
}

Vue.component('vSelect', vSelect)
``` 
Best regards !

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -344,7 +344,7 @@
                 :id="inputId"
                 role="combobox"
                 :aria-expanded="dropdownOpen"
-                aria-label="Search for option"
+                :aria-label="textValues.searchOption"
         >
 
       </div>
@@ -355,7 +355,7 @@
           @click="clearSelection"
           type="button"
           class="clear"
-          title="Clear selection"
+          :title="this.textValues.clearSelection"
         >
           <span aria-hidden="true">&times;</span>
         </button>
@@ -363,7 +363,7 @@
         <i v-if="!noDrop" ref="openIndicator" role="presentation" class="open-indicator"></i>
 
         <slot name="spinner">
-          <div class="spinner" v-show="mutableLoading">Loading...</div>
+          <div class="spinner" v-show="mutableLoading">{{ textValues.loading }}</div>
         </slot>
       </div>
     </div>
@@ -378,7 +378,7 @@
           </a>
         </li>
         <li v-if="!filteredOptions.length" class="no-options" @mousedown.stop="">
-          <slot name="no-options">Sorry, no matching options.</slot>
+          <slot name="no-options">{{ textValues.noResults }}</slot>
         </li>
       </ul>
     </transition>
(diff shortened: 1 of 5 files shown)


sne3ks:
@sagalbot Thanks you a much for your feedback ! Happy to give my contribution to this repository :)

orblazer:
Hello,
Some news about that feature ?

Thanks.

fcpauldiaz:
What happened to this PR ?

Boycce:
@sagalbot Flip, hope you recover fast man, that sucks.

tomdrissen:
Hi @sagalbot Any update on this?

src/components/Select.vue
```diff
@@ -389,9 +389,10 @@
   import pointerScroll from '../mixins/pointerScroll'
   import typeAheadPointer from '../mixins/typeAheadPointer'
   import ajax from '../mixins/ajax'
+  import text from '../mixins/text'
```
kamerat, line 392: ❗ Not tested, but should work in theory unless I have typoed something ❗
- Since [src/mixins/index.js ](https://github.com/sagalbot/vue-select/blob/master/src/mixins/index.js#L5)is already exporting them all, we can just import them at from once place to save space.
```suggestion
  import { ajax, pointerScroll,  text, pointer as typeAheadPointer  } from '../mixins'
```

src/mixins/text.js
```diff
@@ -0,0 +1,48 @@
...
+         *  noResults : "Sorry no matching options."
+         *  loading : "Loading..."
+         *  searchOption : "Search for options"
+         *  clearSelection : "Clear selection"
+         *  
+         *  @type {Object}
+         */
+        texts : {
+            type: Object, 
+            default : () => ({}),
```
kamerat, line 18: I think it would be nice to actually provide the default props here. The current solution can be misleading when debugging with the vue-devtools by instance, `texts` prop magically has default values the user does not see unless diving in the source here.

Suggestion might be to put the variable `defaultTexts` outside of the scope (top of the file as a const), then reference it both in the default function initially, and in the return statement of the `textValues` computed function.

IMO the `textValues `function is not really needed due to the fact that the user can simply spread in their custom props themselves, but ey 👈😎👉 it's also chill not to do it yourself.

kamerat (commented):
Good job! I thought to give my feedback to some of the fields.
Also, in a general note; maybe the prop name could be more specific? - I feel `texts` can mean a lot, maybe something like `i18n`, `locale`, `translation` etc could be more obvious?

Thank you.

src/mixins/text.js
sagalbot: @kamerat whoops, I sent you to the wrong place! Forgot that I had rebuilt this due to the conflicts https://github.com/sagalbot/vue-select/pull/988/files

In that PR, I've implemented it the way that you're describing here.

amaury-hanser:
Hello guys,   
I just wanted to say that this feature would be really nice to have. :)

PR #742, opened 2019-02-01
Clarification of prop usage.

Garry-H:
This little note will help reduce confusion. Props, `placeholder` and `default`, needed specification as they are similar.


Garry-H:
`default` should be `value`

PR #755, opened 2019-02-13
feat(taggable): retain input focus after adding a tag in noDrop

eriknygren:
What
---
- If the `noDrop` prop is set to true, and the user is typing in tags,
always refocus the input after each added element.

Why
---
Because at the moment the user needs to click on the element after
adding each tag, this allows for a more fluid user experience.

### Before:
vue select before

### After:
vue select after


eriknygren:
@sagalbot hahaha, what can I say.. I'm a patient guy, no complaints here 😅

Issue #677, opened 2018-10-10
Dropdown list is hidden (may be conflict with Bootstrap4)


vicnoob:
Im using Light Bootstrap Template (https://www.creative-tim.com/product/light-bootstrap-dashboard) which is using Bootstrap 4. When I try to add v-select into my app, the dropdown of v-select is not working.

![image](https://user-images.githubusercontent.com/17595317/46723825-f01cfd00-cca2-11e8-886a-9621db22fa3f.png)

When I try to remove light-bootstrap from my app, the v-select work fines, but I can not remove it since we all developed base on it. So any idea what's going on?

abinadi:
I have the same issue. I am also using a creative-tim template and the vue-select dropdown does not display.

jonathantisseau:
I'm using bootstrap 4 and had the same issue. The `.dropdown-list` was created but stayed hidden. I corrected that with a simple ccs rule :
`.v-select.open > .dropdown-list { display: block; }`

sreenadh:
The following CSS rules solved my issue.
```
.v-select.dropdown .dropdown-toggle::after {
    content: normal !important;
}
.dropdown .dropdown-menu
{
    transform: scale(1, 1) !important;
}
```

sagalbot:
v3 adjusted all of the class names to be very specific. This should no longer be an issue.

Issue #662, opened 2018-09-26
CSS Conflict with vue-bootstrap/bootstrap4


soichih:
I am using vue-select to replace vue-bootstrap default select box. I believe this issue started when I upgraded vue-bootstrap to 2.0 (which loads bootstrap 4.1.1), but vue-select stying seems to be broken with the version I am using (v2.5.1)

![screenshot from 2018-09-25 23-25-19](https://user-images.githubusercontent.com/923896/46055700-2f801100-c11b-11e8-801f-3300fcda6589.png)

(See the odd caret mark on the right side of the select box, and the height of the select box seems unnecessarily bloated..)

I am loading bootstrap css like so..

```
import 'bootstrap/dist/css/bootstrap.min.css'
import 'bootstrap-vue/dist/bootstrap-vue.min.css'
```

If I remove these, the issue goes away.
![screenshot from 2018-09-25 23-34-11](https://user-images.githubusercontent.com/923896/46055772-8980d680-c11b-11e8-87eb-59fd00909882.png)

I am not sure if this is vue-bootstrap issue, but is there anyway to make vue-select not conflict with vue-bootstrap?

soichih:
I've patched this issue with following CSS rules for now

```css
 /*prevent bootstrap-vue css conflict*/
.v-select.dropdown .dropdown-toggle {
content: inherit;
}
.v-select.dropdown .form-control {
height: inherit;
}

```

bakatz:
Having the same issue - above CSS did not fix it for me (possibly because I'm on multiselect mode). For now, I've downgraded to 2.4.x which works fine but requires some other CSS hacks.

DjeeBay:
In my case the following did the trick. But not sure it is the best way to do it :

`.v-select.dropdown .dropdown-toggle::after {
    display: none;
}`

chrislebaron:
CSS fix did not work for me either, but downgrading to 2.4.0 did the trick.

tranghaviet:
> I've patched this issue with following CSS rules for now
> 
> ```css
>  /*prevent bootstrap-vue css conflict*/
> .v-select.dropdown .dropdown-toggle::after {
> content: inherit;
> }
> .v-select.dropdown .form-control {
> height: inherit;
> }
> ```

this work perfectly, thanks.

sreenadh:
The following CSS rules solved my issue.
```
.v-select.dropdown .dropdown-toggle::after {
    content: normal !important;
}
[...]

Issue #596, opened 2018-07-19
Red border for validation


FrancescoMussi:
Hello.
Sorry if the issue has been raised several times already.

#451, #447

But It doesn't seems to have been properly resolved.

I am not talking about a validation mechanism, but purely about the css required for having red borders.
Is there a class that can be applied to make the border of the select red?
Something like 'invalid', 'is-invalid' ?

Thanks.

JasonHenson:
Yeah it would be nice to have a class or, in my case, proper bootstrap support with correct classes being added to the input. I see example using ".dropdown-toggle" to apply border color to manually, but to match Bootstrap's styling we need to be able to add box-shadow to inputs on focus.

rinu:
I was able to add classes dynamically like you would add a class to any other element. On error case I added the `invalid` class and bootstrap made it have a red border. Pretty simple, doesn't require any special support from `vue-select`.

Extrapolator214:
I encountered the same problem. In case someone needs it, my not-so-elegant solution:

```html
<v-select v-model="transaction.account"
          :options="accountOptions"
          :class="selectClass('account')">
</v-select>
```

```js
methods: {
      state: function (attribute) {
        if(this.validated){
          if(this.errorFor(attribute)){
            return false
          } else {
            return true
          }
        } else {
          return null
        }
      },
      selectClass: function (attribute) {
        let state = this.state(attribute)
        if ( state ) {
          return 'border rounded-lg border-success'
        } else if ( state === false) {
          return 'border rounded-lg border-danger'
        } else {
          return ''
        }
      }
}
```

Issue #535, opened 2018-05-17
CSS Class Collision?


shirshak55:
I guess we can used scoped css so that those css affect components only. If we use tabler admin panel or many other admin panel there is css collision.

So can we think about it 

Thanks :)

PR #759, opened 2019-02-17
Sass & Class Renames

### LINKED ISSUE #574: Rename the `.dropdown` class to be more specific to this component?
This component uses the `.dropdown` class name for the element used to contain the dropdown options. That can be problematic as a `.dropdown` component is pretty generic, and in fact implemented in a lot of UI toolkits. In those cases, the dropdown element generated by this component often inherit (at least some of) the styling rules of those toolkits. And with unexpected/awkward/broken results. Perhaps we could rename this component's `.dropdown` to be something more specific - `v-select-dropdown`, perhaps? Or just kill the entire `.dropdown` class and rename the related classes since we already have the `v-select` class.

I can work up a PR if we can get some guidance on an approach from @sagalbot. Thanks! 👏

sagalbot:
## New

- [x] adds sass-loader
- [x] adds postcss with:
  - [x]  autoprefixer
  - [x]  cssnano 
- [x] all styles have been pulled out of the component and organized into scss modules
- [x] common properties have been extracted into `!default` scss variables prepended with `$vs-`
- [ ] add CSS variables
- [x] spinner slot becomes scoped, receives `loading` boolean prop

## Enhancements

Refined selectors to use a single level of specificity, expect when global component states like `vs--searching` are involved. The refined selectors shave `~2.5kb` off the build, and removes a lot of friction when attempting to override component styles.

## Changes

- [x] `selected-tag` class renamed to `vs__selection`
- [x] all classes have been updated to use the `vs__` prefix
- [x] all CSS selectors have been refactored to use a single selector where ever possible
   ex) `.v-select .selected-tag` becomes `.vs__selection`
- [x] full list of selector changes added to `v3` draft release notes
- [x] remove `dropdown` and `rtl` classes from el
- [x] test default appearance within framemorks:
  - [x] bootstrap 3
  - [x] bootstrap 4
  - [x] Bulma
  - [x] Foundation
- [x] simplify dropdown `ul > li > a` structure by removing unnecessary `a`, refactor all selectors to use `.vs__dropdown-option`
- [x] `dropdownClasses()` renamed to `componentState()`
  - [x] all state classes are prefixed with `vs--`
- [x] **remove `maxHeight` prop**, control via CSS 
- [x] use SVG for deselect and open-indicator
---

Closes #574 #662 #677 #535 #468 

Related #596 #451 #447


docs/.vuepress/components/Sandbox.vue
```diff
@@ -0,0 +1,277 @@
+<template>
+  <div id="sandbox-wrap">
+    <div id="config">
+
+      <div class="list-item" v-if="!hideHelp">
```
eriknygren: Great improvement 👍

src/scss/global/_states.scss
```diff
@@ -0,0 +1,52 @@
...
+ *  RTL - Right to Left Support
+ *
+ *  Because we're using a flexbox layout, the `dir="rtl"`
+ *  HTML attribute does most of the work for us by
+ *  rearranging the child elements visually.
+ */
+
+.v-select[dir="rtl"] {
+  .vs__actions {
+    padding: 0 3px 0 6px;
```
eriknygren: Now when we got postcss and the power of variables, I wonder if it would be a good idea to move some of these magical numbers into variables as well, there's a lot of `6px` padding/marging rules. Could it be worth declaring this once and giving it a suitable name?

src/scss/modules/_clear.scss
```diff
@@ -0,0 +1,15 @@
+/* Clear Button */
+
+$color: rgba(60, 60, 60, 0.5);
+
+.vs__clear {
+  font-size: 23px;
```
eriknygren: same with font-sizes (var it up)

sagalbot: 👍

src/scss/global/_states.scss
sagalbot: Agreed! It's worth considering whether these should be rems or ems as well.

src/scss/modules/_clear.scss
sagalbot: Went a different direction here. The close icons and drop down indicator are now SVGs. I was worried they might bloat the JS a bit, but they were `<1kb`, and let me cut out a whole bunch of CSS, particularly for the drop down indicator. The SVGs scale significantly better and aren't appear consistent regardless of application font settings.

src/scss/global/_states.scss
sagalbot: Leaving this for a `3.x` patch release. It definitely needs to get done but requires a bit more time than I'm willing to give at the moment.

eriknygren:
> @eriknygren I was dragging my feet a bit to finish the last 20% of this today, and your review & comments are a huge motivator! Really appreciate the comments and 👍.

Happy to hear it! Thanks for your effort

PR #808, opened 2019-03-31
Add possibility to change text for empty options 

vmartis:


vmartis:
I need to localize text for empty options.

PR #810, opened 2019-04-02
fixed falsy index values

agm1984:


agm1984:
Let me know if this isn't the branch for for pre-v3 redacts.

agm1984:
I have to retract this. Changing only these parts of code is inadequate. I will try to fix it, but I need a lot more time with the code to understand the true ramifications of such a change.

agm1984:
I was able to fix it again by explicitly checking against `0` and `false`:

``` javascript
        /**
         * Check if there aren't any options selected.
         * @return {Boolean}
         */
        isValueEmpty() {
            const isValueZero = (this.mutableValue === 0);
            const isValueFalse = (this.mutableValue === false);
            if (this.mutableValue || (isValueZero || isValueFalse)) {
                if (typeof this.mutableValue === 'object') {
                    return !Object.keys(this.mutableValue).length;
                }
                return !this.valueAsArray.length;
            }
            return true;
        },

        /**
         * Return the current value in array format.
         * @return {Array}
         */
        valueAsArray() {
            const isValueZero = (this.mutableValue === 0);
            const isValueFalse = (this.mutableValue === false);
            if (this.multiple && (this.mutableValue || (isValueZero || isValueFalse))) {
                return this.mutableValue;
            } else if (this.mutableValue || (isValueZero || isValueFalse)) { // eslint-disable-line no-else-return
                return [].concat(this.mutableValue);
            }
            return [];
        },
```

My current fear is that there are about 32 instances of `this.mutableValue` in the code, and a couple of them look like they could break if a person added `0` or `false` as tags, for example using `pushTags`.

But in my testing, it does seem to work:

image

I'm going to submit the PR again.

agm1984:
I feel a special bond with this code for some reason. I like the type of logic being used in it. I'd like to study the code closer by helping over time :)

I notice there are a few places in the code that use what I would consider fairly complex conditional checks, and in addition to that a couple areas have "concerning" cyclomatic complexity due to `if` blocks within `if` blocks. By concerning, I mean we are splitting hairs here, but I see some opportunities to reduce indentation in a couple areas. I find nested conditionals _can_ spawn difficult to debug anomalies and that this difficulty scales with the number of operators.

I think we could improve odds by making some complex areas slightly more verbose yet simpler to understand.

As a person new to the codebase, I have a hard time understanding how a change to part of a condition such as:

``` javascript
if (val === option || (this.index && val === option[this.index]) || (typeof val === 'object' && val[this.label] === option[this.label])) {
```

would affect the code at run-time because of the number of variables that must be in alignment for that condition to evaluate to true and what that condition really means to which supported use cases. Upon closer inspection, it looks like it's actually simple--looks to be mostly-just checking if the option is a scalar value or an Object. My main issue is the number of operators... and that particular line is 159 characters wide :)

Overall the code is great. My mentions here are mostly stylistic, but I think they could have bearing on minimization of bugs over time and maximization of code-friendliness to 'new contributors' 💯 

I will be happy to make small PRs with specific changes and supplied reasoning, but I would like to a) get v3.0 ASAP and b) start this journey in v3.0 code as it sounds like changes are fairly significant across the board.

agm1984:
Something is still wrong with this PR.

This is the second time my select components went from working to broken:

image

It's almost as if the problem is related to async code, where the extra delay associated with 8 console.logs slows it down enough to work, but when I remove them, it breaks.

agm1984:
Nice, thanks for the reply. I've had this page open for like a week... just circling back now.

I'm still using the library locally with the change as it is in this PR. It has been working fine so far.

I'll be around, can't wait for v3! I'll start looking at the code in forensic detail once that release drops.

Issue #729, opened 2019-01-11
facing issue when doing search operation


Sakthi002:
i removed all the texts in select box but one character is still showing in axios call
![search1](https://user-images.githubusercontent.com/35330401/51033282-e6b86f00-15c8-11e9-9412-0a685788658b.PNG)

Fusseldieb:
That's not a issue @Sakthi002. Those were the previous requests to the server. Why would it search for a empty query?

sagalbot:
Related to #571. Currently the `search` event is not triggered if there is no search query set. 

The proposal in #571 was to add a `search:cleared` event.

sagalbot:
Fixed in 3.0. The `onSearch` callback has been removed in favor of the `@search` event, which is called anytime the string changes, regardless if it is null.

Issue #620, opened 2018-08-22
Remove checking length in select method in ajax mixins


raynox:
```
search() {
			if (this.search.length > 0) {
				this.onSearch(this.search, this.toggleLoading)
        this.$emit('search', this.search, this.toggleLoading)
      }
		},
`````
Can you remove that checking lenght in this method? When you search something, the options change. Then when you remove the whole search text, the options don't change. This is annyoing. And please, reformat that code! :smile: I just copied it from your file.

yordivd:
i agree, does anyone have a workaround for now?

sagalbot:
Fixed in 3.0. The `onSearch` callback has been removed in favor of the `@search` event, which is called anytime the string changes, regardless if it is null.

Issue #571, opened 2018-06-22
ajax mixin should emit `search` event if if text is blank


jacobg:
The ajax mixin has this code:
```
search() {
   if (this.search.length > 0) {
      this.onSearch(this.search, this.toggleLoading)
      this.$emit('search', this.search, this.toggleLoading)
   }
}
```

It seems that there should not be a `length > 0` condition. For example, if a user deletes search text, they'd probably expect to re-query for all results (i.e., without search filter text). If there's cases (such as the demo of github search), where a search text is required, then perhaps there can be a property to toggle that behavior (i.e., allowEmptySearch).

GreymondTheLong:
I need this fixed too, or at least functionality added to recognise when the search value is cleared.

@sagalbot I'm willing to pick this up, I'll just need some guidance and answer to the below questions:

I was looking through the code, and believe we can keep the current functionality but perhaps add a new event for when the search value is cleared. I can see this being useful as people may want to handle clearing search values in a different way to searching (I certainly do).

We could update the maybeDeleteValue function to emit the event when length is 0

sagalbot:
@GreymondTheLong I like the idea of a `search:cleared` event as the `else` of the above conditional branch.

GreymondTheLong:
@sagalbot I'll do that today then. Do you have a contributing guide? Or anything in particular I need to be aware of?

GreymondTheLong:
@sagalbot I've created a pull request #617 

Sorry messed up the commit messages, when I was removing a console log I had forgotten, so there are two commits with same message.

chiptus:
any update on this? or any workaround until it is merged?

kant312:
I'm interested as well! The PR seems to implement this correctly but for some reason the Netlify build did not pass.

GreymondTheLong:
If I get time today, I'll try to fix up the build I did

GreymondTheLong:
Checked it out. It seems like the build configuration for Netlify is missing the build:dev command for all builds

`Executing user command: yarn build:dev
2:33:25 PM: yarn run v1.3.2
2:33:25 PM: error Command "build:dev" not found.`

chiptus:
it seems like this task is missing from `package.json`

sagalbot:
Fixed in 3.0. The `onSearch` callback has been removed in favor of the `@search` event, which is called anytime the string changes, regardless if it is null.

GreymondTheLong:
[...]

Issue #240, opened 2017-06-25
Allow onSearch to query empty string


longprao:
As in the code, the search function only triggers when the input string is not empty

[https://github.com/sagalbot/vue-select/blob/master/src/mixins/ajax.js#L42](https://github.com/sagalbot/vue-select/blob/master/src/mixins/ajax.js#L42)

```javascript
if (this.search.length > 0) {
				this.onSearch(this.search, this.toggleLoading)
        this.$emit('search', this.search, this.toggleLoading)
      }
```

So when I enter a search query, then remove the query, it is stuck with the last letter, so I cannot query for all items anymore, thus no way to disable the search (the onSearch event does not fire at all).

Please remove this check or show a way to remove the entered search. Thanks!

sagalbot:
In hindsight, I can't recall why I added that conditional but I'm sure there was some use-case for it. That said, seems to make way more sense that it'd always run, regardless of wether it was blank. If the end user didn't want that, they could throttle it themselves in their callback.

sagalbot:
![](https://github.trello.services/images/mini-trello-icon.png) [Always emit `search` and trigger `onSearch` even when search text is blank](https://trello.com/c/60U3N9Xi/26-always-emit-search-and-trigger-onsearch-even-when-search-text-is-blank)

sirzerator:
I totally agree that this could be in the component itself. As a workaround, you can always use something like `v-on:keyup.native="someMethod($event)"`.

JRadagast:
As a second workaround, i selected the input and targeted the focus event. I wanted to load the content only when i clicked the element.
```
let inputSearch = this.$refs.elementRef.$refs.search;
inputSearch.addEventListener('focusin', () => this.myFunction());
```

sagalbot:
Fixed in 3.0. The `onSearch` callback has been removed in favor of the `@search` event, which is called anytime the string changes, regardless if it is null.

PR #811, opened 2019-04-02
WIP: v3 – remove onSearch callback prop

### LINKED ISSUE #818: Ajax list not refreshing, when deleting search text at once
You can reproduce the bug on your codepen from the docu:
https://codepen.io/sagalbot/pen/POMeOX

Type a search string into the selectbox like "sagalbo"
image

Then just select the text (for example `CTRL`+`a` or do it with the mouse) and

- delete it  (`DEL` or `BACKSPACE`) 
- OR cut it (`CTRL`+`x` in my system or use context menu from mouse)

and the list isn't getting refreshed:

image

I would expect to get the same result/behaviour like don't selecting the text and just pressing `backspace` until the input is empty:

image

Why is this a bug: Otherwise you could show him (just like on "normal" empty search) a list of the first 10 results.

sagalbot:
The `onSearch` callback is redundant. The `search` event has been in place since 546418da1fc216c72d8b1b31741d4b02f75546b4 /  `v2.1.0` and receives the same parameters: the search string, and `toggleLoading`. The function prop can be removed and in most cases all that will need to change in an implementation is changing `on-search` to `@search`.

- [x] update tests
- [x] update documentation 
- [ ] ~maybe remove `toggleLoading` & `mutableLoading` – defer responsibility of updating `loading` to the client~
   - leaving these properties in place. they don't follow the same pattern as `value` and `options` which must *always* be provided.

---

Closes #818 #729 #620 #571 #240

### DIFF
--- a/src/mixins/ajax.js
+++ b/src/mixins/ajax.js
@@ -19,8 +19,13 @@ export default {
 
   watch: {
     /**
-     * If a callback & search text has been provided,
-     * invoke the onSearch callback.
+     * Anytime the search string changes, emit the
+     * 'search' event. The event is passed with two
+     * parameters: the search string, and a function
+     * that accepts a boolean parameter to toggle the
+     * loading state.
+     *
+     * @emits search
      */
     search () {
       this.$emit('search', this.search, this.toggleLoading);
--- a/src/mixins/ajax.js
+++ b/src/mixins/ajax.js
@@ -7,29 +7,13 @@ export default {
      */
     loading: {
       type: Boolean,
-      default: false
+      default: false,
     },
-
-    /**
-     * Accept a callback function that will be
-     * run when the search text changes.
-     *
-     * loading() accepts a boolean value, and can
-     * be used to toggle a loading class from
-     * the onSearch callback.
-     *
-     * @param {search}  String          Current search text
-     * @param {loading} Function(bool)  Toggle loading class
-     */
-    onSearch: {
-      type: Function,
-      default: function(search, loading) {} // eslint-disable-line no-unused-vars
-    }
   },
 
-  data() {
+  data () {
     return {
-      mutableLoading: false
+      mutableLoading: false,
     };
   },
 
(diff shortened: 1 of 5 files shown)


src/mixins/ajax.js
```diff
@@ -38,20 +22,18 @@ export default {
      * If a callback & search text has been provided,
```
kotarCreative: Maybe consider changing these docs?

src/mixins/ajax.js
```diff
@@ -7,29 +7,13 @@ export default {
...
-     * run when the search text changes.
-     *
-     * loading() accepts a boolean value, and can
-     * be used to toggle a loading class from
-     * the onSearch callback.
-     *
-     * @param {search}  String          Current search text
-     * @param {loading} Function(bool)  Toggle loading class
-     */
-    onSearch: {
```
kotarCreative: Do you think it would make sense to leave this in for now and add a depreciation warning instead?

src/mixins/ajax.js
sagalbot: Good catch, fixed!

src/mixins/ajax.js
sagalbot: Hey @kotarCreative, really appreciate the review and feedback! There's a couple of these callbacks that have been removed for `v3`:

- `onChange`
- `onInput`
- `onSearch`

onSearch in particular is entirely redundant. It receives the same parameters as the event, so in order to adjust an implementation migrating from v2 to v3, `on-search` can be search and replaced to `@search`, and you'll get the same result.

The problem is that if I don't do it now, I'll have to wait until v4 to remove it, and I'd like to get the API consolidated to using events over function props where applicable. I'll backfill the 2.x release with deprecation notices.

Are there any specific use cases you think work better with the function prop vs. the event?

kotarCreative: I think that if the depreciation notices are backed filled it will be ok. I was just concerned about people updating without knowing that this was being removed.

Great work consolidating the event structure 👍

sagalbot: Agreed! I've been maintaining a draft release of v3 with all the breaking changes:

image

I'll also publish an upgrade guide on the new docs at https://vue-select.org.

Appreciate the feedback!!

PR #817, opened 2019-04-05
Validate user input

RayDabbah:
Currently there is no way to validate/filter out a users input when using push tags. This PR adds a prop which is a function to validate a user's input and filter out unwanted content.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -1199,7 +1199,7 @@
       valueAsArray() {
         
         if (this.multiple && this.mutableValue) {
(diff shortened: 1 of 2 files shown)


adityapurwa:
@RayDabbah - I am in need of validating user tags (only number within specific ranges), is there any reason why this PR is closed?

RayDabbah:
@adityapurwa If I remember correctly, there are now merge conflicts with the main branch since the PR was made. I'm not familiar enough with it to fix it since I wrote it so long ago, so I closed it.

Issue #855, opened 2019-04-24
v-model with reduce if value changed outside v-select


fretfan:
**Describe the bug**
When using v-model with reduce and changing v-model value in vue component v-select value is not updated. Input event not emitted.

**To Reproduce**
Pick any value from dropdown
Press reset button
https://codepen.io/anon/pen/PgyLXz
```
<div id="app">
  <h1>Vue Select</h1>
  <v-select v-model="selectedValue" 
            :options="options"
            :reduce="option => option.val">
  </v-select>
  <button @click="resetDropdown">reset</button>
</div>

<script src="https://unpkg.com/vue@latest"></script>
<script src="https://unpkg.com/vue-select@latest"></script>
```
```
Vue.component('v-select', VueSelect.VueSelect)

new Vue({
  el: '#app',
  data: {
      selectedValue: null,
      options: [
        { label: "one", val: 1 },
        { label: "two", val: 2 },
        { label: "three", val: 3 }
      ]
  },
   methods: {
    resetDropdown() {
      this.selectedValue = null;
    }
  },
  updated() {
    console.log(this.selectedValue);
  }
})
```

**Expected behavior**
V-select value is changed with associated v-model variable.
input event emitted

sagalbot:
Thanks for reporting and providing a repro link. I'll get a patch out for this one.

sagalbot:
Forked: https://codepen.io/anon/pen/zXMNRj#anon-login

Notes:
- I think a possible solution is to add a watcher to `selectedValue` computed prop

sagalbot:
Duplicate #842

edwardsph:
Do you have a workaround for this issue whilst you are working on a patch?
Thanks

huypn:
I have tried to update in code and working
Just add this function in watch
```
watch:{
      value(){
        this.updateValue(this.findOptionFromReducedValue(this.value))
      }
}
```
That's all!

Hope #sagalbol help to update it.

Cryde:
> Forked: https://codepen.io/anon/pen/zXMNRj#anon-login
> 
> Notes:
> 
>     * I think a possible solution is to add a watcher to `selectedValue` computed prop

I've try to do that but with no success.  
Any comment on the @huypn patch ?

wazum:
Thanks @huypn 
> I have tried to update in code and working
> Just add this function in watch
> 
> ```
> watch:{
>       value(){
[...]

PR #889, opened 2019-06-06
Update once value changed outside

huypn:
This change will help to update select component once value changed outside follow issue #855


Slimmmo:
This works perfectly for single selects but crashes on multiple selects.

Issue #895, opened 2019-06-21
CSS selector for single v__select span


lucatrioschi:
I'd like to apply css style to the `span.vs__selected` element in `:multiple="true"` v-select.
The slot template will not do, because it's contained in the `span`, and i want to apply my style (border and background color) to the whole `span` depending on the selected value.
Example:
Options list: ['Ok', 'No', 'I will think about']
Selected options: ['Ok', 'No']
Current behaviour: all `span.vs__selected` are gray and i can control content with template slots
What i'd like: a way to select a single span

My 2 cents:
```
// Select.vue - line 16
<span class="vs__selected" v-bind:key="option.index" :ref-value="option.value">
  <!-- unchanged -->
</span>
```
This way i can css select span per span, and have a red span for "No" option and green span for "Ok" option.
```
span.vs__selected[ref-value="Ok"]{
  background-color: green;
  border: 1px solid darkgreen;
}
span.vs__selected[ref-value="No"]{
  background-color: red;
  border: 1px solid darkred;
}
/* ... */
```

Issue #894, opened 2019-06-21
Deselect currently selected option on click (multiple select)


lucatrioschi:
When using a `:multiple="true"` v-select, deselect the clicked option if already selected.
Example:
Options list: `['a', 'b', 'c']`
Current selection: `['a', 'c']`
Action: click on the v-select (opens the dropdown) and click on opion 'a'
Current behaviour: nothing happends
Requested behaviour: deselect option 'a' (new current selection: `['c']` )

Maybe this behaviour can be enabled/disabled via prop (false by default, so compatibility wont break).

My 2cents:
```
// Select.vue - line 480 
select(option) {
  if (!this.isOptionSelected(option)) {
    if (this.taggable && !this.optionExists(option)) {
      option = this.createOption(option)
    }
    if (this.multiple) {
      option = this.selectedValue.concat(option)
    }
    this.updateValue(option);
  } else { // my 2 cents here
    if (this.multiple && this.disableSelectedOptionOnClick){ // new prop, pls change name! :D
      this.deselect(option);
    }
  }
  this.onAfterSelect(option)
},
```

Thanks

PR #877, opened 2019-05-08
Adding deselectable prop to enable deselceting options when selected an option is selected again

telliott22:
Implementing the suggestion from https://github.com/sagalbot/vue-select/issues/561#issuecomment-411624684

PR #896, opened 2019-06-21
Update Select.vue

lucatrioschi:
see issue
https://github.com/sagalbot/vue-select/issues/895
and
https://github.com/sagalbot/vue-select/issues/894


andreasvirkus:
@sagalbot If you have the time, please consider reviewing this. We'd really appreciate the feature to deselect via an option

WofWca:
The second change is the same as in #877

PR #898, opened 2019-06-24
Add label to new taggable item in dropdown

Turbotailz:
This adds a label beside new options in the drop-down list to help distinguish it from other options. Example:

Screenshot from 2019-06-24 17-22-36

The text can be changed by setting the prop `new-option-label` to any string. It can be disabled by setting the string to `""`.


msonowal:
@sagalbot 
This is a must have when push tag is used, We also update the UI to have plus sign for example at before.
So imo, I think one approach would be having a new css class to be applied for the tags that is created or can be created

MohamedAliTera:
Any chance this gets merged?

Issue #790, opened 2019-03-13
Dropdown doesn't toggle when clicking selected option within a scoped-slot


dev-yarolsav:
![image](https://user-images.githubusercontent.com/6161950/54293263-4d9dd880-45b8-11e9-9f55-f7fc9141ce36.png)

Only clicking on whitespace triggers dropdown.

sagalbot:
This happens when using the scoped slots, because the click handler is no longer applied to the node.

`toggleDropDown` checks for clicks on certain elements, in this case it's looking for `.selected-tag`, but the target is actually the inner div. This method needs to be refactored:

https://github.com/sagalbot/vue-select/blob/bda47806d3dc5112d1838083aa4f397c14929ee9/src/components/Select.vue#L901-L920

StefanoGroenland:
I am having the same issue, @sagalbot when will this be fixed?

sagalbot:
@StefanoGroenland too hard to give you an estimate. Would gladly accept a PR with a fix.

elhowm:
I have the same issue. Is there some hotfix solution to avoid this behavior?

alxgb:
> I have the same issue. Is there some hotfix solution to avoid this behavior?

The only workaround (short of fixing the library internally) is adding a `vs__selected ` class on **every** subchild of an overwritten `selected-option` slot. ` vs__selected` has styling though, so you may need some CSS hacks to prevent it from applying.

wrabit:
Hi, any progress on this?

sagalbot:
FYI this was fixed in 3.3.

PR #904, opened 2019-07-10
Fix dropdown not opening on overriden selected-option slots (#790)

alxgb:
Even though #869 is open it seems to be a larger change, and as far as I can tell it's not really fixed, the only commit is a cleanup of toggleDropdown. Since the issue is tagged "help-wanted" anyway, I'm opening this

This fix (for #790) is fairly simple, instead of checking for vs__selected we check whether the target is a child node of the selectedOptions ref. I then realized that with this, clicking the deselect button would open the dropdown, which felt a bit weird, so I added the bit of extra logic to exclude those.


wrabit:
Please merge 🙏🙏✌️

PR #914, opened 2019-07-27
Make sure selected value is an option after option changed and react to value property changes even if tracking value internally

doits:
Before this case did not work correctly:

- Select was rendered with *no* options, but *with* a saved value
- Options were fetched by ajax and options prop was updated
- Reduce function is passed

What happens without this commit is that the selected tracked value
simply was the raw reduced value (previously saved). Which means that
displaying a label for example does not work if the label comes from the
unreduced option.

This commit makes sure that the internal tracked value is checked
against all options not only once the select is created but additionally
when options change.

---

To test this in your project, you can use [my custom branch](https://github.com/doits/vue-select/tree/fix_initial_label_on_ajax_options_built) which incoperates this PR and #921 and has built packages.

**Disclaimer**: Not for production use, only provided for testing purposes. Might be bugged and eat your data.

In your `package.json`:

```json
"dependencies": {
    "vue-select": "https://github.com/doits/vue-select.git#fix_initial_label_on_ajax_options_built"
}
```

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -8,13 +8,12 @@
 
       <div class="vs__selected-options" ref="selectedOptions">
         <slot v-for="option in selectedValue"
-              :key="getOptionKey(option)"
               name="selected-option-container"
               :option="normalizeOptionForSlot(option)"
               :deselect="deselect"
               :multiple="multiple"
               :disabled="disabled">
-          <span class="vs__selected">
+          <span :key="getOptionKey(option)" class="vs__selected">
             <slot name="selected-option" v-bind="normalizeOptionForSlot(option)">
               {{ getOptionLabel(option) }}
             </slot>
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -13,7 +13,7 @@
               :deselect="deselect"
               :multiple="multiple"
               :disabled="disabled">
-          <span class="vs__selected" v-bind:key="option.index">
+          <span class="vs__selected">
             <slot name="selected-option" v-bind="normalizeOptionForSlot(option)">
               {{ getOptionLabel(option) }}
             </slot>
@@ -55,7 +55,6 @@
         <li
           role="option"
           v-for="(option, index) in filteredOptions"
-          :key="index"
           class="vs__dropdown-option"
           :class="{ 'vs__dropdown-option--selected': isOptionSelected(option), 'vs__dropdown-option--highlight': index === typeAheadPointer }"
           @mouseover="typeAheadPointer = index"
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -437,11 +437,17 @@
       /**
        * Maybe reset the value
        * when options change.
+       * Make sure selected option
+       * is correct.
        * @return {[type]} [description]
        */
       options(val) {
         if (!this.taggable && this.resetOnOptionsChange) {
           this.clearSelection()
+        }
+
+        if (this.$options.propsData.hasOwnProperty('reduce') && this.value) {
+          this.setValueFromOptions()
         }
       },
 
@@ -460,17 +466,25 @@
       this.mutableLoading = this.loading;
 
       if (this.$options.propsData.hasOwnProperty('reduce') && this.value) {
-        if (Array.isArray(this.value)) {
-          this.$data._value = this.value.map(value => this.findOptionFromReducedValue(value));
-        } else {
-          this.$data._value = this.findOptionFromReducedValue(this.value);
-        }
+        this.setValueFromOptions()
       }
 
       this.$on('option:created', this.maybePushTag)
     },
 
     methods: {
+      /**
+       * Make sure tracked value is
+       * one option if possible.
+       * @return {void}
+       */
+      setValueFromOptions() {
+        if (Array.isArray(this.value)) {
+          this.$data._value = this.value.map(value => this.findOptionFromReducedValue(value));
+        } else {
+          this.$data._value = this.findOptionFromReducedValue(this.value);
+        }
+      },
 
       /**
        * Select a given option.
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -445,8 +445,18 @@
           this.clearSelection()
         }
 
-        if (this.$options.propsData.hasOwnProperty('reduce') && this.value) {
-          this.setValueFromOptions()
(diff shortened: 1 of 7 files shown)


Kocal:
I can confirm that this PR fix the issue with inital value's label and XHR-loaded options.

**Before:**
Sélection_999(071)

**After: :tada:**
Sélection_999(072)

> Not for production use, only provided for testing purposes. Might be bugged and eat your data.

Well, this is definitly a must have feature/fix and since we have a project to ship... it will be used on production :smile:

robertmarney:
#925 is corrected by this PR

ipetrov87:
This fix a problem that I am trying to find a solution from two days. Hopes it will be merged soon. :)

whiteglasses:
Need this fix very much.

geri777:
I have tried the version from _doits:fix_initial_label_on_ajax_options_ and I have the problem that the displayed/selected value changes from the label to the id immediatley after the selection. I am using the reduce option.

doits:
@geri777 do you have any example code?

geri777:
@doits I could not reproduce it in my sample - and finally found out that I was changing the model of the list when I selected a value. The new list had no corresponding element to the selected value and thus the record ID was displayed. My fault... sorry for the confusion.

geri777:
Here is my demo: https://plnkr.co/edit/ITy2eHO4gc1JU8UjIrmG?p=preview
But again, this is not a problem related directly to your improvement. Your improvement is highly appreciated!

doits:
@geri777 here's a fork with what I wrote in #933: https://next.plnkr.co/edit/AOsN7N4nkQEIfyzX which at least fixes the display problem and does not load an empty list. This might be a good work around in the meantime.

doits:
@geri777 see https://next.plnkr.co/edit/rUVy1dr5O26pdHbt for another update where it always returns the option for the current value, irregardless of the search value. This should be a complete work around in the meantime.

geri777:
@doits - hmm - you have removed the part from my example where the option list is manipulated. This is not a workaround. I know my example is a bit misleading - but what I want to demonstrate is a change of the server-side result, based on the search value. 
Your idea of forcing the server to return the currently selected value in the option list - regardless of search parameter is the only workaround which I think is valid, but not usable in the daily business.
I think this issue has to be handled in vue-select. Most likely we need a further internal data object for the selected value which is decoupled from the option list.

doits:
I don't think I removed the part where it changed the options in that fiddle but regardless of it I too think the selected option should be cached for that case. When I find some time I can try to implement it and see how it goes, but currently I am a little bit overloaded.

src/components/Select.vue
```diff
@@ -13,7 +13,7 @@
               :deselect="deselect"
               :multiple="multiple"
               :disabled="disabled">
-          <span class="vs__selected" v-bind:key="option.index">
```
sagalbot: Here and `:58`, why drop the `key` binding?

src/components/Select.vue
```diff
@@ -55,7 +55,6 @@
         <li
           role="option"
           v-for="(option, index) in filteredOptions"
-          :key="index"
```
sagalbot: Same as above.

src/components/Select.vue
doits: In this case `option.index` was not defined at all if I understood this correctly, since `option` can be anything (a string, object, number, ...) - whatever the developer passes. Please correct me if I am wrong.

src/components/Select.vue
doits: Here it is simply using `index` as the key. It is defined but since it is the `index` of the `filteredOptions` it does not stick to the option. This means the same option might get a different `index` when `filteredOptions` changes.

If I understand the purpose of `:key` correctly it should be chosen so that it sticks to the thing you are iterating over, in this case `option`. If you render the same `option`, it should always have the same `:key`, so the dom node will be moved around and must not rerendered completely.

All in all I don't think it makes any real difference here though, because the template is so simple and there are no transitions or other special vue logic here. But if there were, having the `:key=index` might break it.

Most correctly there should be a `:key` for every `option`, but I don't know how to derive it correctly from every `option` format.

src/components/Select.vue
sagalbot: Ahh yes, totally correct!

src/components/Select.vue
sagalbot: Great catch here. What do you think of deriving the `key` using `JSON.stringify(option)`. The only complication I can think of is if you had duplicate options, but I think you'd be encountering a bunch of bugs in that case anyways.

doits: I think `JSON.stringify` actually might work in 99 % of all cases. Out of my head it will break only in the special case if `option` is an object with a circular reference (you see I assume people pass the weirdest things as option ;-)). Should be tested though.

If this is implemented overrideable as a prop though (e.g. `getOptionId` similar to `getOptionLabel`), then it will work in 99 % of all cases and can be overriden if it fails.

Maybe it should by default look if it is an object and has an `id` property and if yes use it (like with `.label`), otherwise fall back to `JSON.stringify`. Something like

```js
props: {
  getOptionId: {
    type: Function,
    default(option) => {
     if (typeof option === "object" && option.id) {
        return option.id
      } else {
        return JSON.stringify(option)
      }
    }
  }
}
```

sagalbot: Yeah, I like that a lot. 

Not sure if I prefer `getOptionKey` or `getOptionId`. `getOptionKey` is a bit more explicit.

doits: I added it right now. No tests yet though, do we need them?

sirber:
Thank you for fixing this issue! Can't wait for a release!

doits:
> I tried the current 914.patch and the issue of v-model updated in the parent not changing selection is still there

Are you sure? You used my branch from the OP? If there is still an issue, please create a fiddle so I can hunt this down.

sirber:
I applied the patch made by GitHub of this pull request on the master
branch. Sorry I already switched to another select component.

jackmcdade:
Hello! Is there anything I can do to get this one merged in? Do we need to increase the code coverage?

jackmcdade:
Awesome!

doits:
@sagalbot glad to see it in the wild! Hope it does not have too many bugs ;-) Glad I could help.

PR #309, opened 2017-09-06
Support option with disabled property

markselby9:
1, Add support for options with disabled property
2, Add corresponding test case and demo

issue reference: https://github.com/sagalbot/vue-select/issues/104

PR #921, opened 2019-07-31
Allow to disable options with selectable function

doits:
Implemented as per https://github.com/sagalbot/vue-select/pull/309#issuecomment-476777951

ToDo:

- [x] Write spec
- [x] Have it tested by some people

---

To test this in your project, you can use [my custom branch](https://github.com/doits/vue-select/tree/fix_initial_label_on_ajax_options_built) which incoperates this PR and #914 and has built packages.

**Disclaimer**: Not for production use, only provided for testing purposes. Might be bugged and eat your data.

In your `package.json`:

```json
"dependencies": {
    "vue-select": "https://github.com/doits/vue-select.git#fix_initial_label_on_ajax_options_built"
}
```

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -57,9 +57,9 @@
           v-for="(option, index) in filteredOptions"
           :key="index"
           class="vs__dropdown-option"
-          :class="{ 'vs__dropdown-option--selected': isOptionSelected(option), 'vs__dropdown-option--highlight': index === typeAheadPointer }"
+          :class="{ 'vs__dropdown-option--selected': isOptionSelected(option), 'vs__dropdown-option--highlight': index === typeAheadPointer, 'vs__dropdown-option--disabled': !selectable(option) }"
           @mouseover="typeAheadPointer = index"
-          @mousedown.prevent.stop="select(option)"
+          @mousedown.prevent.stop="selectable(option) ? select(option) : null"
         >
           <slot name="option" v-bind="normalizeOptionForSlot(option)">
             {{ getOptionLabel(option) }}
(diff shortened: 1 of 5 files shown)


tobyzerner:
Works well, thanks for your work! A few issues:

* Unselectable options can still be navigated to with the arrow keys
* Unselectable options can still be selected by pressing Enter
* Unselectable options still appear in filtered results

doits:
Thank for the PR @tobyzerner and fixing those remaining issues in https://github.com/doits/vue-select/pull/1 - I'll respond here about the business logic so we have everything in one place about it.

I agree with everything but I'm not sure if disabled options should be removed when filtering looking from a business logic stand point. Because actually it might be good to show the user the the disabled option so the user knows there is that option but it is simply not selectable. Maybe because it is temporary disabled or because of some precondition. Otherwise he might think this option does not exist at all.

Of course this depends on what you use disabled options for. Maybe you have something in mind where showing disabled options when filtering makes no sense. Can you maybe give an example use case?

tobyzerner:
True, that's a very good point. My use case for disabled options is actually to simulate optgroups which is why I don't want them to appear in filtering. So it's probably not a good idea to include that logic by default – I've removed it and rebased my PR. For my particular use-case, I realised I can use the `filter`/`filterBy` props to prevent them from showing up!

doits:
> I think once [doits#1](https://github.com/doits/vue-select/pull/1) is merged this one is good to go as well. Thanks so much for the great work!!

Alright, I merged it, should be ready to go now.

doits:
@sagalbot feel free to merge and release this too, if everything is well here :-)

Issue #923, opened 2019-08-06
Select is not updating when v-model was changed


tikagnus:
**Describe the bug**
Select is not updated when the `v-model` variable was changed, at least when we have a custom `reduce` and object type options.

**To Reproduce**
1. Declare a variable `entity` in data()
2. Use `entity` as `v-model` or `:value`
3. Update `entity` in a method of current component
4. Select was not updated

**Expected behavior**
Select should be updated when we change the `entity` value.

**Screenshots**
![image](https://user-images.githubusercontent.com/8859382/62513196-cd586800-b823-11e9-8310-3cf42c887ca9.png)

**Desktop (please complete the following information):**
 - OS: MacOs Mojave
 - Browser Chrome
 - Version 75

**Additional context**
```
<template>
  <v-select
    :reduce="item => item.id"
    label="label"
    :options="entities"
    v-model="entity"
  >
  </v-select>
</template>

<script>
  export default {
    name: 'Dropdown',
    data () {
      return {
        entity: null,
        entities: []
      }
    },
    methods: {
      async handleSearch (search, loading) {
        loading(true)
        try {
          const res = await this.$http.get('persons', {
            params: {
              q: search
            }
          })
          const { data } = res
          this.items = data
        } catch (e) {
        }

        loading(false)
      },     
    },
    beforeMount () {
      this.entities = [
        {
          id: 1,
          label: 'a'
        }, {
          id: 2,
          label: 'b'
        },
      ]
      setTimeout(e => {
        console.log('entity value before change', this.entity)

        this.entity = 2
        console.log('entity value after change', this.entity)
      }, 2000)

    }
  }
</script>
```

This issues was not in 2.6 version.

As a temporary solution I added a watcher on the `value` https://github.com/tikagnus/vue-select/commit/360b0b427d0d984c3eb4b1e4b82a70f289917507

doits:
Should be fixed by #914

sagalbot:
Fixed by #914.

PR #924, opened 2019-08-06
Added a watch for value.

btica-ro:
https://github.com/sagalbot/vue-select/issues/923

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -434,6 +434,12 @@
     },
 
     watch: {
+      value(newVal){
+        const option = this.findOptionFromReducedValue(newVal)
+        if(option){
+          this.updateValue(option)
+        }
+      },
       /**
        * Maybe reset the value
        * when options change.


doits:
See #914 for my approach (only handling it when value is tracked internally).

agrism:
To do not change directly in the package, you can extend vue-select by creating your component and using your component.
```
<script>
    import vSelect from 'vue-select';

    export default {
        extends: vSelect,
        watch: {
            ...vSelect.$watch,
            value(newVal) {
                const option = this.findOptionFromReducedValue(newVal);
                // if (option) {
                    this.updateValue(option);
                // }
            }
        }
    }
</script>

**Comment by agrism** [comment]
to not make changes in the package you can create own component and extend vue-select 
```
<script>
    import vSelect from 'vue-select';

    export default {
        extends: vSelect,
        watch: {
            ...vSelect.$watch,
            value(newVal) {
                const option = this.findOptionFromReducedValue(newVal);
                if (option) {
                    this.updateValue(option);
                } else {
                    this.updateValue('');
                }
            }
        }
    }
</script>
```

PR #935, opened 2019-08-30
Change all keyup events to keydown

tobyzerner:
For a few reasons:

- event.preventDefault() for the Enter key (to stop it from submitting
  the form when you select an item) is only effective if it's a keydown
  event.

- Using keydown for up/down navigation means you can hold them down to
  rapidly scroll through a lot of items.

- Keydown events make the UX feel more responsive, and is consistent
  with how most apps/operating systems handle key presses.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -781,16 +781,10 @@
           case 9:
             //  tab
             return this.onTab();
-        }
-      },
-
-      /**
-       * Search 'input' KeyBoardEvent handler.
-       * @param e {KeyboardEvent}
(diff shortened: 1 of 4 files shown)


ziming:
Hi @sagalbot, it is not fully fixed yet. There are still issues with pressing enter after you typed in the vue select field.

If you have a **required** text field **before** the vue select field and you type in the vue select field **1st** and press enter, the 1st required text field will tell you the 1st field is required and the vue select value goes blank.

This does not happen in vue select 2.

Tested it in google chrome

Issue #661, opened 2018-09-25
Tags and case sensitivity


ravisangar27:
I'm using a vue-select with taggable="true"  and I'm trying to prevent adding a new tag for "foo" when "Foo" already exists as an option. 

What I'm trying to accomplish is to check, that if I do already have a "Foo" option available, it should use that one and not add a new "foo", which should be the same as the already existing "Foo".

mikedidthis:
At the moment it is possible for a user to select both `one` and `One`. 

I would like all `values` to default to `lowercase` (`one`,`two`, `three`). If the user types `one` or `One` the value `one` if selected.

Here is a codepen showing the incorrect behaviour: https://codepen.io/mikedidthis/pen/ErJvRr

Any help would be appreciated.

sagalbot:
https://codepen.io/sagalbot/pen/aroLdp?editors=1010

```html
  <v-select 
    :options="optionsString"       
    :filterable="false"
    :multiple="true"
    :taggable="true"
    :create-option="option => option.toLowerCase()">
  </v-select>
```

☝️ that will solve your problem when options are primitives – although it'd break if you'd enter a 1, so you'd probably want a more robust `create-option` callback.

PR #951, opened 2019-09-29
Add 'compare' props

vq0599:
Fix issue similar #661 .

- default:
```js
// default
comparea(label, search) {
  return label === search
}
```
image

- with comparea
```js
//  ignore case 
comparea(search, label) => {
    return search.toLowerCase() === label.toLowerCase()
  }
```
image

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -309,6 +309,19 @@
       },
 
       /**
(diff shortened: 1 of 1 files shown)


vq0599:
Fixed by #1012

Issue #956, opened 2019-10-04
Allow additional keyboard keys to create new tag when using taggable feature


peteyb:
**Is your feature request related to a problem? Please describe.**
We use the taggable feature to allow our users to enter email addresses to forward invitations to our site to their colleagues. Before moving to Vue.js and vue-select we used jQuery UI which allowed us to configure more than just the Enter key to create new tags. 

**Describe the solution you'd like**
As email addresses do not contain spaces we would like our users to be able to hit the Space key to trigger new tags being created, not just the enter key.

**Describe alternatives you've considered**
I am reluctant to move back to jQuery within our vue spa

**Additional context**
Can you add a way of specifying a list of which keys trigger a new tag creation, defaulting to current Enter

khamer:
I created a PR for this functionality above. Let me know if this would provide the functionality you need.

peteyb:
@khamer That would be great, should work a treat.

Meekohi:
Nice timing. I'm very interested in this.

sagalbot:
@khamer your PR is pretty close to how I imagined solving this problem. The challenge is that folks have different requirements for how the want the component to respond to input of all types, so I'd like to abstract it a bit higher.

#956 adds a new prop: `handlers`. It can be used to overwrite default functionality, or add handlers for different keys. It maps keycodes to a callback function.

If I have a taggable input, and I want `,` `return` or `space` to 'tag' the current option, you could use `handlers` to make that happen.

```html
<v-select taggable multiple no-drop :handlers="handlers" />
```

```js
export default {
  methods: {
    handlers(handlers, vm) {
      const tagging = e => {
        e.preventDefault();
        vm.typeAheadSelect();
        vm.searchEl.focus();
      };

      handlers.search[32] = tagging;
      handlers.search[188] = tagging;
      handlers.search[9] = tagging;
      handlers.search[13] = tagging;

      return handlers;
    }
  }
}
```

This API does require a bit more knowledge of component inner-workings, but also provides a whole lot of flexibility for all types of use cases. What do you think of this solution?

Meekohi:
I think you mean #971 ^

Personally I think it'd be great to merge #971 as the general purpose mechanism and then consider some "short hand" props for common use cases that just add the correct key-map under the hood.

sagalbot:
thanks @Meekohi – updated to point to the right PR. I'll investigate some shorthand syntax today.

khamer:
[...]

PR #958, opened 2019-10-07
Add selectOnCustom to override select keys

khamer:
This adds a new prop, `selectOnCustom`, where users can override the default key for triggering selection or add multiple keys. For example, For Issue #956, you could use `:select-on-custom="[13, 32]"`.

Let me know if there's other changes you'd like to see. Things that come to mind are

- We could likely simplify/deprecate `selectOnTab` to combine it with this functionality.
- I wonder if there's a reason `selectOnTab` is based on keydown, whereas all other bindings are based on keyup? Is that something we'd need to allow `selectOnCustom` to specify?

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -454,6 +454,15 @@
       searchInputQuerySelector: {
         type: String,
         default: '[type=search]'
+      },
+
+      /**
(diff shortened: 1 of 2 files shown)


Meekohi:
Can close this I think -- was added to #971 and merged.

PR #973, opened 2019-10-26
Fix 969 add get selected option class

### LINKED ISSUE #969: Individual class for selected options on multiple select with selected-option-container
Hi, I'm trying to apply individual class for every option so after I can give them different styles with CSS.

I've copied&pasted your example and put it inside a v-select multiple like this:

`<v-select :options="options" multiple>
    <slot v-for="option in options" name="selected-option-container" :option="(typeof option === 'object')?option:{[label]: option}" :deselect="deselect" :multiple="multiple" :disabled="disabled">
        <span class="selected-tag" :class="'unique-tag--'+option.id" v-bind:key="option.index">
            <slot name="selected-option" v-bind="(typeof option === 'object')?option:{[label]: option}">
                {{ option }}
            </slot>
            <button v-if="multiple" :disabled="disabled" @click="deselect(option)" type="button" class="close" aria-label="Remove option">
                 <span aria-hidden="true">&times;</span>
             </button>
    </span>
 </slot>
</v-select>`

I've a really simple script:

`var app = new Vue({            
    el: '#app',
    data: {
        options: [
            {id: '1', label: 'Option green'},
            {id: '2', label: 'Option blue'},
            {id: '3', label: 'Option pink'},
            {id: '4', label: 'Option yellow'},
            {id: '5', label: 'Option purple'}
        ]
    },
    methods: {}
})`

And it isn't working. I only receive a lot of messages like these:
[Vue warn]: Property or method "multiple" is not defined on the instance but referenced during render.
[Vue warn]: Property or method "deselect" is not defined on the instance but referenced during render.
[Vue warn]: Property or method "disabled" is not defined on the instance but referenced during render.

Any help? 
Please!

Thank you so much in advance.

sagalbot:
It's currently really difficult to add a custom class to the `selected-option-container` slot. That slot needs to be refactored, and this prop provides a lightweight solution to custom classes for selected options in the meantime.

---

Closes #969

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -13,7 +13,7 @@
               :deselect="deselect"
               :multiple="multiple"
               :disabled="disabled">
-          <span :key="getOptionKey(option)" class="vs__selected">
+          <span :key="getOptionKey(option)" :class="getSelectedOptionClasses(option)">
             <slot name="selected-option" v-bind="normalizeOptionForSlot(option)">
               {{ getOptionLabel(option) }}
             </slot>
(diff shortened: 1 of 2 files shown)


McPo:
Running into this issue now, Is there any update to this draft feature?

PR #988, opened 2019-11-09
Improved Internationalization

sagalbot:
Based off the work in #741. Rebuilding here to avoid some pretty hectic conflicts.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -17,7 +17,7 @@
             <slot name="selected-option" v-bind="normalizeOptionForSlot(option)">
               {{ getOptionLabel(option) }}
             </slot>
-            <button v-if="multiple" :disabled="disabled" @click="deselect(option)" type="button" class="vs__deselect" aria-label="Deselect option">
+            <button v-if="multiple" :disabled="disabled" @click="deselect(option)" type="button" class="vs__deselect" :aria-label="i18n.deselect.ariaLabel">
               <component :is="childComponents.Deselect" />
             </button>
           </span>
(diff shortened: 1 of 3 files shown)


skipworkgh:
What's the status on this PR?

jja-financial-com:
Any updates for this PR?

src/components/Select.vue
```diff
@@ -83,6 +83,7 @@
   import pointerScroll from '../mixins/pointerScroll'
   import typeAheadPointer from '../mixins/typeAheadPointer'
   import ajax from '../mixins/ajax'
+  import i18n from '../mixins/i18n'
```
kamerat, line 86: I would suggest this, but this is purely subjective as it more or less does the same thing..

```suggestion
   import { ajax, pointerScroll, i18n, pointer as typeAheadPointer  } from '../mixins'
```

## Notice ‼
However, seems you have not added i18n to the exports in [src/mixins/index.js](https://github.com/sagalbot/vue-select/blob/master/src/mixins/index.js ), making my proposal above not working.

kamerat (approved):
Really clean approach 💯🤩

bcartier:
Will this PR be reconsidered for future releases of vue-select?

docs/.vuepress/components/TranslationSpanish.vue
```diff
@@ -0,0 +1,17 @@
...
+
+SpanishSelect.props.locale = {
+  type: Function,
+  default: () => spanish,
+}
+
+export default {
+  components: { SpanishSelect },
+}
+</script>
```
sagalbot, line 17: Last thing to do is to fix this example - it gets overridden by the default vue select component registration.

PR #1002, opened 2019-11-16
Call filter callback regardless of search query

### LINKED ISSUE #1001: Proposal: `filter` should always be applied regardless of search length
In certain situations, such as working with a large dataset, the consumer requires fine grained control over the options list. `filter` is currently the best way to hook into `filteredOptions` and control the output. However, `filter` is only applied to the options *if* `this.search.length` is true. What if the consumer needs to control the output even if the search string is null?

https://github.com/sagalbot/vue-select/blob/22bf8f9475762adc0a1f49b6483c2840be186c7a/src/components/Select.vue#L1069-L1081

Line 1076 should be refactored:

```js
+ let options = this.search.length ? this.filter(optionList, this.search, this) : optionList;
- let options = this.filter(optionList, this);
```

- the filter callback is always run, regardless of search length
   - [ ] _is this a breaking change?_
- instead of passing `search` *and* `this`, just pass `this`
   - the consumer can destructure `filter(options, { search })`

**Additional context**
Related to #240 #508 - emitting `@search` with empty search string. API should have parity with this feature.

sagalbot:
As described in #1001.

# Breaking Change
`filter` is called with an empty search string

---

Closes #1001

### DIFF
--- a/docs/.vuepress/components/AirportCodes.vue
+++ b/docs/.vuepress/components/AirportCodes.vue
@@ -1,0 +1,116 @@
+<template>
+  <div class="flex">
+    <div class="airport-select default">
+      <table>
+        <tr>
+          <th colspan="2">Without Optimization 😬</th>
+        </tr>
+        <tr>
+          <td>Dropdown Render</td>
+          <td :class="{ slow: !! timing.default }">{{ timing.default }}ms</td>
+        </tr>
+      </table>
+      <observed-select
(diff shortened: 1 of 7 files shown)


marianheinsen:
Whats the status of this PR? I need to filter even if there is no search input because I want to show the dropdown only if the user searched for something.

gubisim:
Any update on this?

PR #1012, opened 2019-11-29
Compare Options with getOptionKey instead of label + reduce

### LINKED ISSUE #80: Duplicate labels different ids
Hi,
I have a list of objects which contain value and label. If label is the same but the value is different If you select one of them It considers both selected.

### LINKED ISSUE #804: Cant select 2 options with them same name
Hello everyone , i have options (Companies) with the same name ex Adidas ,but with different id and i need to select both, because first company is Adidas Germany and second is Adidas France how can i do this?
this is my code :   
         <v-select multiple   
         label="name" 
         v-model="selectedCompany"  
         :filterable="true"
          :options="companies" 
         @search="onSearch">

### LINKED ISSUE #912: Binded model dosn't update after second select [SOLVED]
The model is not updated after option select ones
**To Reproduce**
[Code Pen Example](https://codepen.io/porcelanosa/pen/OKRagL?editors=1011)
**Expected behavior**
The model should change when the option is selected if the option is already selected.

I temporarily solved my problem by change component code by adding 
```js
methods: {
     select: function(t) {
    this.clearSelection(); // HERE!
   this.isOptionSelected(t) || (this.taggable && !this.optionExists(t) && (t = this.createOption(t)), this.multiple && (t = this.selectedValue.concat(t)), this.updateValue(t)), this.onAfterSelect(t);
	},```

### LINKED ISSUE #970: Using `reduce` + slots + option objects missing default label prop without providing a `label` prevents vue-select from selecting a new value
**Describe the bug**
If you are using `reduce` in combination with slots and you do not specify a `label` prop, then vue-select won't fire `@input` event when a new option is selected.

**To Reproduce**

https://codepen.io/owenconti/pen/XWWRzEG

**Expected behavior**

Selecting a new option even without `label` should fire the `@input` event for the selected option.

 - OS: Mac
 - Browser: Chrome
 - Version: 77

**Additional context**

Couple things I think are happening here:

1. `console.warn` is no longer fired because vue-select provides the dist build which has stripped console statements:

```
        terserOptions: {
          compress: {
            drop_console: true,
          },
        },
```

So we would have probably caught this right away if console statements were displayed, however I still think requiring a label is not necessary when using reduce + slots.

2. Essentially this bug happens because the of the order of operations in `optionComparator`:

```
        if (value === this.reduce(option)) {
          return true;
        }
        if (
          this.getOptionLabel(value) === this.getOptionLabel(option) ||
          this.getOptionLabel(value) === option
        ) {
          return true;
        }
        if (this.reduce(value) === this.reduce(option)) {
          return true;
        }
```

The first `if` statement fails (as expected), however the second `if` statement passes because `this.getOptionLabel()` returns `undefined` when no `label` is provided and your option objects do not contain the default `label` value of "label".

I think if the order of operations changed such that both reduce `if` statements were run before the `getOptionLabel` check, then `label` could be ignored entirely for reduce + slot usage.

### LINKED ISSUE #1007: Won't trigger change listener if the option label is the same
e.g i've option object looks like this

`option:[{code:1,label:'Halo'},
{
code:2,label:'Halo'
}]`

it wont trigger any change events, i've tried @input, :on-change , binding with v-model, reducing it,  nothing will fire it. is it a bug or what?

### LINKED ISSUE #1045: Input event fires only if label changed
At my project I have list of options that can have equal labels, but different values.
Here is example https://codepen.io/petrukhnov/pen/PowXPEz

I think it's better to fire 'input' event on value change.

### PR DESCRIPTION
**Pull request by sagalbot**
This change updates option comparison functions so that the options generated key is used for comparison instead of the label. 

---

Fixes #80 
Fixes #804 
Fixes #970 
Fixes #912
Fixes #1007
Fixes #1045

### CODE REVIEW

File: src/components/Select.vue [hunk 1]
```diff
@@ -693,25 +693,7 @@
...
-          if ((this.getOptionLabel(value) === this.getOptionLabel(option)) || (this.getOptionLabel(value) === option)) {
-            return true
-          }
-          if (this.reduce(value) === this.reduce(option)) {
-            return true
-          }
-        }
-
-        return false;
+        return this.getOptionKey(value) === this.getOptionKey(option);
```
sagalbot: - [ ] determine if the changes to this method introduce any breaking changes

sagalbot: nope, this resolves a bug – not a breaking change.

PR #1014, opened 2019-11-29
Prevent resetting when the options change to the same object

jasonvarga:
Hi!

When using the `resetOnOptionsChange` option, the value is reset even if the new options object contains the exact same set of items.

We are populating the options via an AJAX request, so technically yes they are different objects, but the actual options never changed.

This PR makes it so that it checks that it's actually changed.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -513,8 +513,10 @@
        * is correct.
        * @return {[type]} [description]
        */
-      options(val) {
-        if (!this.taggable && this.resetOnOptionsChange) {
+      options(val, oldVal) {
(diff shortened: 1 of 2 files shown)


jasonvarga:
Ah yeah, fair questions.

I feel like `resetOnOptionsChange` implies that "the options" themselves are changing, not just "the JS object you passed to the `options` prop" changing. In which case I'd say it's a bug fix.

I wouldn't be surprised if someone disagreed.

Regarding the checking for value behavior, I think that feels even better.

jasonvarga:
Good idea. Works for me!

jasonvarga:
Legend. 💪

PR #1036, opened 2020-01-07
Display custom label in place of selected options

### LINKED ISSUE #1003: Show ‘3 items selected’ instead of the actual 3 tags
The problem is that when you have a search box which includes several drop downs you need to make sure the layout does not break and  Vue-select will NOT grow in height to fit all the selected items/tags.

The perfect solution is to show the number of items selected istead on the actual tags which also showing the clear icon of the right to clear all options. Or even better a slot allowing us to control how the selected tags are rendered.

**Describe alternatives you've considered**
snapappointments/bootstrap-select

**Additional context**

PNG image

andreasvirkus:
This PR resolves #1003 with an additional prop, whose textual value would override the `vs__selected-options` contents.

Maybe `collapseTags` would be a better name, seems that Element UI uses that as mentioned in https://github.com/sagalbot/vue-select/issues/1003#issuecomment-567502356

An example use case for this new prop would be:
```
<v-select
  v-model="userIds"
  label="name"
  placeholder="Everyone"
  :options="users"
  :customFieldLabel="userIds.length > 2 ? `${userIds.length} selected` : null"
  :reduce="option => option.id"
  :closeOnSelect="false"
  @search="getUsers"
/>
```

custom-label

### DIFF
--- a/docs/api/props.md
+++ b/docs/api/props.md
@@ -48,6 +48,17 @@ clearable: {
 },
 ```
 
+## customFieldLabel
+
+Custom field value to display in case of long or numerous values. Useful when you want to display the dropdown inline and can't have it wrap multiple lines.
+
+```js
+customFieldLabel: {
+	type: String,
+	default: ''
+},
+```
+
 ## maxHeight
 
 ::: warning Deprecated in `v2.x` & Removed in `v3.0`
@@ -340,10 +351,10 @@ createOption: {
 ## resetOnOptionsChange
 
 When false, updating the options will not reset the selected value.
- 
-Since `v3.4+` the prop accepts either a `boolean` or `function` that returns a `boolean`. 
 
-If defined as a function, it will receive the params listed below. 
+Since `v3.4+` the prop accepts either a `boolean` or `function` that returns a `boolean`.
+
+If defined as a function, it will receive the params listed below.
 
 ```js
 /**
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -7,21 +7,24 @@
     <div ref="toggle" @mousedown.prevent="toggleDropdown" class="vs__dropdown-toggle">
 
       <div class="vs__selected-options" ref="selectedOptions">
-        <slot v-for="option in selectedValue"
-              name="selected-option-container"
-              :option="normalizeOptionForSlot(option)"
-              :deselect="deselect"
-              :multiple="multiple"
-              :disabled="disabled">
-          <span :key="getOptionKey(option)" class="vs__selected">
-            <slot name="selected-option" v-bind="normalizeOptionForSlot(option)">
-              {{ getOptionLabel(option) }}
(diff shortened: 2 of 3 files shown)


andreasvirkus:
@sagalbot could this be reviewed please? 🙏 We're really looking forward to getting back to using the new versions of `vue-select`

dekadentno:
Is there a way to fix and merge this? Sorry for being pushy, I just hope this marvelops feature isn't forgotten. Also thank you for your time and effort on this lib @andreasvirkus @sagalbot

andreasvirkus:
ping @sagalbot

adamdyderski:
I'm afraid that it will not be possible to deselect the selected options when the tags will not be rendered? (clicking on an option in the list does not deselect it now). Correct me if I'm wrong

andreasvirkus:
@adamdyderski I think the functionality you're looking for is in a different PR - https://github.com/sagalbot/vue-select/pull/1033

This just adds the functionality to show a custom label, so it's only a visual change in that sense. It would greatly benefit from https://github.com/sagalbot/vue-select/pull/1033 of course, but I've decoupled the features into 2 separate PRs

docs/api/props.md
```diff
@@ -48,6 +48,17 @@ clearable: {
 },
 ```
 
+## customFieldLabel
+
+Custom field value to display in case of long or numerous values. Useful when you want to display the dropdown inline and can't have it wrap multiple lines.
+
+```js
+customFieldLabel: {
+	type: String,
```
sagalbot, line 57: I'm torn on whether this should be a string or a function (or both). If it was a function, it could be called from a computed property, passing along the currently selected values. Something like this:

```vue
props: {
  shouldTruncate: {
    type: Boolean,
    default: false
  },
  truncatedLabel: {
    type: [String, Function],
  }
},
computed: {
  truncatedLabelResult() {
    if( typeof this.truncatedLabel === 'function') {
      return this.truncatedLabel(this.value)
    }

    return this.truncatedLabel;
  }
}
```

You can easily make the string a computed property from the parent component, but people always seem to struggle with this. In my experience, I almost always wrap `VueSelect` in a parent component to handle that sort of thing. In a perfect world, the component wouldn't manage any state - only firing events, it'd be up to the developer to track it themselves.

Either way, I think we do need some sort of `shouldTruncate` flag that can be referenced from the template. That way you can truncate if more than X values are selected. Maybe it could be a slot? With a slot, we wouldn't need `shouldTruncate`, we can just check if the slot has any content, and if it does, we use that content instead of iterating the options. 

Let me know if you've got any thoughts on the above! Thanks for all your work on this.

gautam-patadiya:
ping @sagalbot

gautam-patadiya:
new Limit property added:
https://github.com/sagalbot/vue-select/issues/1559#issuecomment-1978789674

PR #1047, opened 2020-01-23
feat: add dropRenderingMode prop

WofWca:
Use cases:
* The options list is heavy and it is better to render it once rather than on each open.
* The user wants the width of the field to be no more than the width of the widest option, so it does not change when another option is selected.
This can be achieved by keeping the dropdown in the DOM at all times and keeping its `position` `static`.

### DIFF
--- a/docs/api/props.md
+++ b/docs/api/props.md
@@ -369,6 +369,20 @@ noDrop: {
 },
 ```
 
+## dropRenderingMode
+
+Determines rendering mode for the dropdown ('v-if' or 'v-show').
+
+```js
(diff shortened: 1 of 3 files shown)


WofWca:
I'm having second thoughts about this change, starring package size.

WofWca:
I'm also thinking that the goal can be achieved by putting the `v-if="dropdownOpen"` part inside one of the slots, so the user can control things like `<transition>`s, `<keep-alive>`s, `v-if`s and `v-show`s.

PR #1048, opened 2020-01-24
* Prevent Escape key from propagating and potentially closing modals …

ahfeel:
…for example


andreasvirkus:
I think this is better handled user-side and not within the component, no? Can't you solve it currently with:

```
<modal @keyup.native.esc="closeModal">
  <vue-select @keyup.native.stop />
</modal>
```

With this PR, we'd block people from using Vue's handy event modifiers with event listeners inside templates and would force them to always pass in the handler via a prop.

PR #1049, opened 2020-01-24
* Add append-to-body directive and autoFocus option.

ahfeel:
* Add two new options:
- appendToBody: Boolean default false, this option allows to append the dropdown menu at the end of the body. In this case the sizing and positioning of the menu is done in javascript.
- autoFocus: automatically put the focus on the search input when the component is mounted


zanedev:
LGTM would love to see this merged asap @sagalbot  👍

src/components/Select.vue
```diff
@@ -500,6 +503,26 @@
...
+      appendToBody: {
+        type: Boolean,
+        default: false
+      },
+
+      /**
+       * Focus the input when the component is mounted.
+       * @type {Boolean}
+       */
+      autoFocus: {
```
sagalbot: Updating this signature to `autofocus` brings it in-line with the HTML5 attribute, little easier to remember.

src/components/Select.vue
```diff
@@ -566,6 +589,14 @@
       this.$on('option:created', this.maybePushTag)
     },
 
+    mounted() {
+      if (this.autoFocus) {
+        this.$nextTick(() => {
+          this.searchEl.focus();
+        });
+      }
+    },
```
sagalbot: Instead of applying focus programmatically, can you bind the `autofocus` attribute to the search input? This way we can leverage standards and defer the focusing to the browser.

ahfeel: Unfortunately this does not work as the html5 autofocus attribute only works on page load, so we still have to do it programmatically :)

src/components/Select.vue
ahfeel: done !

ahfeel:
Also, I tried to add unit tests on these two options but couldn't find a way to get a behavior close enough to reality..

src/directives/appendToBody.js
```diff
@@ -0,0 +1,19 @@
+export default {
+    inserted: function (el, bindings, vnode) {
+        if (vnode.context.appendToBody) {
+            let rect = vnode.context.$refs.toggle.getBoundingClientRect();
+
+            el.style.width = rect.width + 'px';
```
sagalbot: I'd love to pull the positioning logic here into a `positioning` prop from the component. I've been looking for a way to include popper.js to handle the dropdown positioning logic and a pluggable function here just might be the answer.

PR #1060, opened 2020-02-07
Allows users to control the behavior of clearSearchOnBlur.

diasbruno:


package.json
```diff
@@ -9,7 +9,7 @@
     "test": "tests"
   },
   "private": false,
-  "main": "dist/vue-select.js",
+  "main": "./src/index.js",
```
sagalbot: Will need this reverted to merge.

src/components/Select.vue
```diff
@@ -1038,7 +1047,7 @@
        * @return {Boolean} True when single and clearSearchOnSelect
        */
       clearSearchOnBlur() {
-        return this.clearSearchOnSelect && !this.multiple
+        return !this.dontClearSearchOnBlur && this.clearSearchOnSelect && !this.multiple
```
sagalbot: Rather than add a new prop, how about pulling `clearSearchOnBlur` up to a function prop instead? This way the default stays the same, and the user gets even more control over the implementation.

package.json
diasbruno: My bad. Included this patch by mistake.

src/components/Select.vue
diasbruno: I was trying to figure out what to expose that doesn't require changes to the current API.
I'll see what I can do.

diasbruno: Done.

src/components/Select.vue
```diff
@@ -425,6 +425,17 @@
         validator: (value) => ['function', 'boolean'].includes(typeof value)
       },
 
+      /**
+       * If search text should clear on blur
+       * @return {Boolean} True when single and clearSearchOnSelect
+       */
+      clearSearchOnBlur: {
+        type: Function,
+        default: function () {
```
sagalbot: Looks great! I think the last thing we can do to clean it up is pass `clearSearchOnSelect` and `multiple` as arguments. That way the consumer has access to them in their implementation. Probably best passed as an object.

diasbruno: Done.

diasbruno:
Ping. Also, let me know if I can help reviewing some PRs and issues.

PR #1068, opened 2020-02-18
Add and update WAI-ARIA 1.1 combobox propeties

### LINKED ISSUE #757: Missing `aria-owns` attribute
According to the ARIA spec, use of the `role="combobox"` requires some ARIA attributes that this component is missing. 

First it looks like you're implement the older Aria 1.0 pattern.  So a larger fix would be to update it to the 1.1 pattern (see more info in link below)

Using the 1.0 pattern (putting `role="combobox"` on the input element) you're missing the `aria-owns` attribute, and of course give the popup `<ul>` an id.  This is probably where I would let users pass in an ID prop themselves so they can handle uniqueness and you don't have to deal with it.

source: https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-6

> In a combobox implementing the ARIA 1.0 pattern:
> - The element that serves as the textbox has role combobox.
> - **When the combobox popup is visible, the element with role combobox has aria-owns set to a value that refers to an element with role listbox.**
> - the element with role combobox has a value for aria-haspopup of listbox. Note that elements with role combobox have an implicit aria-haspopup value of listbox.

### LINKED ISSUE #866: WAI-ARIA Combobox - Complete aria- attributes
The `aria-explanded` & `combobox` attributes should be moved up to the `$refs.toggle` node. Currently they are set on the search input, but the toggle node houses information relevant to the combobox. 

Example from the spec:

```html
<div aria-label="Tag" role="combobox" aria-expanded="true" aria-owns="owned_listbox" aria-haspopup="listbox">
    <input type="text" aria-autocomplete="list" aria-controls="owned_listbox" aria-activedescendant="selected_option">
</div>
<ul role="listbox" id="owned_listbox">
    <li role="option">Zebra</li>
    <li role="option" id="selected_option">Zoom</li>
</ul>
```

- [ ] pull up `role=combobox`
- [ ] pull up `aria-expanded`
- [ ] add all missing `aria-` tags

### LINKED ISSUE #1027: Aria-expanded should be set to "false" when the dropdown is not open
**Describe the bug**
According to the accesibility standards. The "aria-expanded" should be set to "false" when the dropdown is not open.

**To Reproduce**
1. Go to https://vue-select.org/ and see the selector's HTML in the console log. For your convenience here it is:
image

**Expected behavior**
The element should contain the attribute `aria-expanded="false"` before it is even opened. (Of course, it should go to "true" when opened and back to "false" when closed again)

 - OS: Any
 - Browser: All

### LINKED ISSUE #1072: Causes an error on WAVE accessibility report
**Describe the bug**
When a vue-select items is viewed in the browser and a WAVE report is run, an accessibility issue is flagged. 

This is important because our institution (along with many others) determine the accessibility status of an application/page based on the results of a WAVE report.

WAVE is indicating that the "Clear selection" button doesn't have an accessible name or containing text, so a screen reader user cannot read or access the button.

**To Reproduce**
1. Install [WAVE Browser Plugin](https://wave.webaim.org/extension/)
2. Go to a page in a browser that has the vue-select like the [vue-select homepage](https://vue-select.org/)
3. Run WAVE report
4. Look in WAVE results under "Errors"

**Expected behavior**
When WAVE is run, no errors are indicated

**Screenshots**
See attached
Screen Shot 2020-02-27 at 3 37 29 PM

 - macOS (10.14.6)
 - Chrome (but may be an issue in others)
 - Version 80.0.3987.122

geo-connor:
This branch adds several aria properties outlined in the WAI-ARIA combobox spec:
https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-6
With further inspiration taken from here:
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

I've tested the rendered component with the NVDA screen reader, and support there has been improved.

Since many of the element relationships outlined in the spec depend on id attributes being set, I have added those attributes. To keep the values unique, I have included an int created by lodash's `uniqueId` when the component instance is created (ex: `vs3__combobox`).

The w3.org example renders the `ul` component even when the list is empty. I can see how this would be necessary for some screen readers to establish the `aria-owns` relationship. I was able to do this while (I believe) avoiding the issues described in #609 regarding v-show/v-if. Basically, the `ul` will have a `v-show`, and the list items will not render thanks to a `<template v-if>`.

`aria-expanded="false"` will render now when the listbox is closed (#1027). This just needed a `.toString()` since vue doesn't render some bound attributes that equal `false`, `null`, `undefined`, etc.

---

Related #609
Resolves #757
Resolves #866
Resolves #1027
Resolves #1072


src/components/Select.vue
```diff
@@ -505,6 +510,7 @@
 
     data() {
       return {
+        uid: uniqueId(),
```
sagalbot: I think we can get away with using `this._uid`. It's a private component property and the implementation may change in the future, but should work for connecting the dots within the component. I've used it quite a bit for label/id relations in my own apps and have never run into issues.

Related: https://github.com/vuejs/vue/issues/5886#issuecomment-308625735

geo-connor: I think that would work here too. It's my understanding that using `this._uid` is not a recommended practice:
https://github.com/vuejs/vue/issues/5886#issuecomment-308625735

But I have used it in other projects and have not had any issues yet.

sagalbot: @geo-connor I ended up writing the function myself. Turns out it's a super simple implementation 👍

geo-connor: Looks great!

PR #1083, opened 2020-03-05
feat: scope the no-options slot

### LINKED ISSUE #1071: Feature: Add new option if there is no search match
### **Is your feature request related to a problem? Please describe.**
**Use case:** User searches for something that is not currently in the list of options, but he would like to add it.

**Problem**
Currently, the slot `no-options` does not have any slot-scope options. Therefore, it is not possible to get the current search value to use for adding a new option.

### **Describe the solution you'd like**
1. The search input as slot-scope data for `no-options`
2. (optional) Built in feature to add not found option to options & select it.

The .2 is the thing I am trying to achieve, but without slot-scope data, I am not sure how to obtain this functionality. 

### **Describe alternatives you've considered**
I have not thought about any other way to solve this within the component.

### **Additional context**
Idea of what I'm trying to achieve:
Screenshot 2020-02-20 at 16 50 02
`something` in cursive there is hardcoded, but would ideally come from a slot-scope

### LINKED ISSUE #1081: No options text shown on empty input
**Describe the bug**
When i focus empty v-select input, i see `no-options` item.
Maybe i do something wrong?

**To Reproduce**
I use these settings:
```
<v-select v-model="user"
  @search="fetchUsers"
  :filterable="false"
  :options="users"
  placeholder="User name"
  :reduce="(user) => { return { name: user.name, guid: user.guid } }" 
  label="name"></v-select>
```

**Expected behavior**
I expect only focused element 
or maybe some helper text "Enter username or choose from options..."

sagalbot:
Resolves #1071, Resolves #1081

### DIFF
--- a/docs/.vuepress/components/BetterNoOptions.vue
+++ b/docs/.vuepress/components/BetterNoOptions.vue
@@ -1,0 +1,16 @@
+<template>
+  <v-select>
+    <template v-slot:no-options="{ search, searching }">
+      <template v-if="searching">
+        No results found for <em>{{ search }}</em>.
+      </template>
+      <em style="opacity: 0.5;" v-else>Start typing to search for a country.</em>
+    </template>
+  </v-select>
+</template>
+
+<script>
+export default {
+  name: 'BetterNoOptions',
+};
+</script>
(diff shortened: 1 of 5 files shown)


BlitZz96:
💯 Good job!

PR #1092, opened 2020-03-09
new option - enableMouseSearchInput

dkurucz:
**Summary:**
Adds a configurable option to allow mouse input events on any area of the search input, rather than closing the dropdown outright.

**Purpose of this feature:**
There are instances where users may want to click somewhere in the text they have typed in order to select it and do any number of operations, including but not limited to copying and pasting, deleting certain parts of longer words in order to modify the search string, or to double-click-select an entire word in the search string, or triple-click-select and highlight the entire search string.

Note that this feature is additionally complementary to the other option I have put in another MR, called preserveSelectedText (https://github.com/sagalbot/vue-select/pull/1093), which will automatically populate a single-selection (not multiple) vue select, allowing users to click on choices that were previously chosen and improving speed to enter additional or modified text, and improves usability overall.

This feature is flag-enabled and does not break any default functionality.

Please let me know if there is anything needed to improve naming conventions or if I missed any edge cases, thanks!


dkurucz:
@sagalbot any thoughts on getting this in as well as #1093?  I ask because my project is currently using a fork that has the functionality in place.  Let me know if you need me to do anything else with this work.  Thanks!

Dares300:
Hello everybody,

I'm 100% agreed with Jeff. It would be genial if we could have this feature. It is the default behavior expected by anyone.
It is very disturbing to have the possibility to do that inside a normal input and not just bellow in a vue-select. The normal user does'nt care about the difference between a normal input or select. Event if it is lot of work to do (I can understand because vue-select is already complex but awesome !).

Thank you for this library Sagalbot and thank you for your very details mail Jeff  !

Have a good week !

Dawid HILLIEN
A fan of vue-select in Paris

________________________________
De : Jeff Sagal <notifications@github.com>
Envoyé : jeudi 19 mars 2020 17:27
À : sagalbot/vue-select <vue-select@noreply.github.com>
Cc : Dares300 <hillien2@hotmail.com>; Manual <manual@noreply.github.com>
Objet : Re: [sagalbot/vue-select] new option - enableMouseSearchInput (#1092)


@dkurucz<https://github.com/dkurucz> just had a chance to check it out today. My first instinct is that this should actually be default behavior.

This feels like a bug to me:

[Mar-19-2020 10-20-54]<https://user-images.githubusercontent.com/692538/77095662-76c32700-69cb-11ea-905c-d60e13a2931b.gif>

And with enableMouseSearchInput:

[Mar-19-2020 10-21-36]<https://user-images.githubusercontent.com/692538/77095741-8c385100-69cb-11ea-9f1f-f49d76dd16db.gif>

I can't think of a use case where enableMouseSearchInput should be false. You can still click the caret to close (although it should probably have cursor: pointer on hover). I think chances of people finding the right prop to enable this behavior is also fairly low.

Do you agree that enableMouseSearchInput should just be default behavior? If so, lets drop the prop and implement it as such, and I'll push it out as a patch release.

________________________________

This is a bit related to what I was working on in #1067<https://github.com/sagalbot/vue-select/pull/1067>.

dkurucz:
@sagalbot yes I agree with your thoughts on this.  I think it might be better to have the opposite as an option, where it would close on click only if the user is expected to pick an option discreetly.  There is a case in my project right now where that could be the preferred behavior.  I would equate this to a "choose one of the following" type of input, where you also wouldn't even want the user to type to search.  However, at this point it would be a nice to have.

If you'd like, I'll take another pass and get it working by default and then put up a proposal for the flag (or maybe a PR if I can get it all in one go).

dkurucz:
@sagalbot I have removed the flag and made the behavior default, feel free to re-test, etc, I think this is largely good to go.  Let me know if you have other thoughts on this.  Thanks!

dkurucz:
@sagalbot any thoughts on a timeline to get this feature in? thanks!

MoogyG:
This functionality break the use of a router-link in "selected-option" slot like this : 
```
<template v-slot:selected-option="customerScope">
    <router-link class="mr-2" :to="{ name: 'customer', params: { id: customerScope.id } }">
        {{ customerScope.name }}
    </router-link>
</template>
```
This shouldn't be enabled by default.

dkurucz:
@MoogyG I'll look into the router link issue when I have a moment.  In regards to your other concerns, this issue was accepted as default behavior (it was originally coded as an option), as a result no option was included or configured to enable/disable this behavior, so there will be no additions to the documentation about it.  The unit test was an oversight, and should be removed since we are no longer testing the option.

dkurucz:
@MoogyG I tested and fixed your above concerns in the following PR: https://github.com/sagalbot/vue-select/pull/1211.  Feel free to test on your end as well.

MoogyG:
I just get your PR, build it localy, copy "dist" folder in my node_modules but it's not working on my machine, I still can't click on a router link inside vue-select.

dkurucz:
@MoogyG can you verify you also copied the compiled css in addition to the built js into the dist directory of node_modules vue-select?  Both are required in that fix.  If so, can you send me a screenshot of what you are seeing?  Let's also move this discussion to the new PR.  Thanks!

MoogyG:
My bad, I was importing SCSS from `src` folder, not CSS from `dist` folder.
Now it works, thanks for your support!

PR #1093, opened 2020-03-09
new option - preserveSelectedText

dkurucz:
Summary:
Adds a configurable option to allow the search text to automatically be populated with the selected value when the dropdown is opened.

Purpose of this feature:
There are instances where users may want to immediately re-select the value or perform another search by taking off a letter at the end of the selected item, for example, or any number of additional reasons the user just wants to interact more directly with the value that they chose (or pre-populated in the data model).  Note, this feature is complementary to https://github.com/sagalbot/vue-select/pull/1092 and with both options enabled allows for much more fluid interactions with the control.

This feature is flag-enabled and does not break any default functionality.

Please let me know if there is anything needed to improve naming conventions or if I missed any edge cases, thanks!

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -485,6 +485,15 @@
       },
 
       /**
+       * Enables/disables updating the search text to the last selected value when the dropdown is opened.
+       * @type {Boolean}
+       */
+      preserveSelectedText: {
+        type: Boolean,
+        default: false
+      },
+
+    /**
        * Query Selector used to find the search input
        * when the 'search' scoped slot is used.
        *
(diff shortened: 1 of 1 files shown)


src/components/Select.vue
```diff
@@ -904,6 +913,10 @@
       onSearchFocus() {
         this.open = true
         this.$emit('search:focus')
+
+        if (!this.multiple && this.preserveSelectedText) {
+          this.search = (this.value || {})[this.label] || ''
```
sagalbot: I think this may be a little more robust:

```js
this.search = this.selectedValue.length ? this.getOptionLabel(this.selectedValue[0]) : '';
```

dkurucz: @sagalbot you are correct, I have made the change, seems to be working well

dkurucz:
@sagalbot any updates on getting this feature in? thanks!

PR #1122, opened 2020-03-20
Check before accessing offsetHeight.

perdjurner:
I think I've found a bug with using the "selectable" prop. If you start typing text that only matches an option that has been set to not selectable, then press tab, I get this error: "Cannot read property 'offsetHeight' of undefined".

My fix just puts an if statement around reading the "offsetHeight" property.

### DIFF
--- a/src/mixins/pointerScroll.js
+++ b/src/mixins/pointerScroll.js
@@ -32,8 +32,10 @@ export default {
       let pixelsToPointerTop = 0;
       if (this.$refs.dropdownMenu && this.dropdownOpen) {
         for (let i = 0; i < this.typeAheadPointer; i++) {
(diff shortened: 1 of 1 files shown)


perdjurner:
@sagalbot I see that there are now conflicts that need to be resolved before merging. Was this issue fixed in another code commit perhaps? If so, this pull request can just be closed I guess :)

mix3d:
pointerScroll was updated in v10 due to adding in a new prop. I'd double check that your issue is still reproducible in the current code; the PR could still be helpful if the issue is still occurring.

perdjurner:
I tried the new release and I think my issue is now fixed.

Thanks!

PR #1158, opened 2020-04-12
Fix/create option bug

### LINKED ISSUE #1156: `createOption` is not creating an object
**Describe the bug**
When using `createOption` in 3.9.5 like this:
```
                                <v-select
                                    label="name"
                                    :options="foo"
                                    v-model="organization"
                                    taggable
                                    :create-option="org => ({ name: org})"
                                    id="newOrg"
                                >
```

`organization` is getting saved as a string and not an object.
**To Reproduce**
1. Try a vue-select:
```
                                <v-select
                                    label="name"
                                    :options="foo"
                                    v-model="organization"
                                    taggable
                                    :create-option="org => ({ name: org})"
                                    id="newOrg"
                                >
```
2. See the value of `organization` when you create a new one.

**Expected behavior**
```
organization: {
  name: "new org"
```

sagalbot:
Closes #1156

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -695,7 +695,7 @@
        */
       onAfterSelect(option) {
         if (this.closeOnSelect) {
-          this.open = !this.open
+          this.open = !this.open;
           this.searchEl.blur()
         }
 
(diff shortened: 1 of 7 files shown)


src/components/Select.vue
```diff
@@ -1153,10 +1153,13 @@
         }
 
         let options = this.search.length ? this.filter(optionList, this.search, this) : optionList;
-        if (this.taggable && this.search.length && !this.optionExists(this.createOption(this.search))) {
-          options.unshift(this.search)
+        if (this.taggable && this.search.length) {
+          const createdOption = this.createOption(this.search);
+          if (!this.optionExists(createdOption)) {
+            options.unshift(createdOption);
```
sagalbot, line 1159: This was the bug. The search string was being added to the list instead of the result of `createOption`.

PR #1162, opened 2020-04-16
New option - optionKey, also fixes the highlight on previously selected options

dkurucz:
Purpose of this PR:
1. allows the specification of an optionKey directly to substitute the need to write a comparator function override for getOptionKey.  This allows developers to take advantage of the existing getOptionKey default function where many sanity checks on options already exist.
2. adds a missing highlight style for tagged items and adds a check to not hover-highlight options already selected.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -61,7 +61,7 @@
           :key="getOptionKey(option)"
           :id="`vs${uid}__option-${index}`"
           class="vs__dropdown-option"
-          :class="{ 'vs__dropdown-option--selected': isOptionSelected(option), 'vs__dropdown-option--highlight': index === typeAheadPointer, 'vs__dropdown-option--disabled': !selectable(option) }"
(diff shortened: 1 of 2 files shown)


mauriciojsola:
Is this going to highlight the item that is already selected?
At this point, for example, if you have 200 item in the select and you have, lets say, 150 selected options, you see all those tagged, but if you make a search, you see all the option (even the ones you already selected). There is no way to distinguish among the non-selected options with the selected one.
That would be awesome to have an option to add a checkbox to each option too.... but that is too much for now :)

dkurucz:
@mauriciojsola what you described would be the same behavior if you were to override the getOptionKey function and specify the comparator in a similar way, so yes it would highlight all options (visible based on the search) if the property you specified was not unique to the list item.  This PR allows you to accomplish this and other use cases, but through a simpler, more direct specification.

@sagalbot any thoughts on getting this in?  I am currently using a fork of this repo to support this need in a production environment (including the styling enhancement).  Is there anything you need from me?  Thanks!

Issue #884, opened 2019-05-28
Highlighting is different to the behavior of a normal select element


simonbrent:
If you give vue-select a value prop, I would expect that value to be highlighted in the dropdown the first time it is opened.

In fact, I think the highlighting behavior is confusing in general. A normal <select> element highlights the current value when opened, and then the focused/hovered option as you interact with the list.

vue-select uses the typeAheadPointer value to remember the last focused/hovered option, meaning that that option is highlighted again when the dropdown is reopened. This does not produce the same behavior as a normal select, since you can close the dropdown by clicking on an element outside the vue-select, without updating the vue-select's value.

I propose that when the dropdown is opened, typeAheadPointer is set to the index of the option which matches selectedValue[0], if selectedValue is an array of length 1.

This behavior could be turned on/off with a prop, if necessary.

mroustchev:
Is there a workaround for achiving something similar with the current version?
This behaviour is critical for the usability when working with longer option lists

ocomsoft:
I am not sure if this warrants  PR but I fixed this by adding the following watch to the typeAheadPointer mixin

`
open(newValue) {
      if (newValue === true) {
        // Just Opened dropdown

        for (let i = 0; i < this.filteredOptions.length; i++) {
          if (this.isOptionSelected(this.filteredOptions[i])) {
            // Highlight the first selected option
            this.typeAheadPointer = i;
            this.$nextTick(function() {
              // Scroll to it if there is a lot of options. 
              this.maybeAdjustScroll();
            });
            break;
          }
        }
      }
    }
`

dorian-marchal:
As a workaround, I created a wrapper component that can be used instead of `vue-select` to fix this behavior:

```javascript
<template>
  <VueSelect v-bind="$attrs" @open="onOpen" v-on="$listeners" ref="select" />
</template>

<script>
import VueSelect from 'vue-select'

export default {
  name: 'Select',
  components: { VueSelect },
  methods: {
    onOpen(...args) {
      if (this.$refs?.select) {
        // Fix a bug in vue-select where selected index is not the default one on open.
        const select = this.$refs.select
        const reduce = select.reduce ?? (_ => _)
        const optionIndex = (select.options ?? []).findIndex(option => reduce(option) === select.value)
        select.typeAheadPointer = optionIndex
        this.$nextTick(function() {
          select.maybeAdjustScroll?.();
        });
      }

      if (this.$listeners.open) {
        this.$listeners.open(...args)
      }
[...]

Issue #1169, opened 2020-04-20
Default value set using v-model is not selected in the list


mattWalters0:
I am setting a default value with v-model but this model is not actually selected in the list when you open the dropdown:

https://codepen.io/MattWalters/pen/XWmjKpM

If I manually select the value then it works ok.  But how I can I get the default value to show as selected?

sagalbot:
The default you've provided is "selected" in both cases. Let me confirm the UX that you're looking for:

- dropdown scrolls to the selected option when opening
- the selected option has a blue background

mattWalters0:
Hi, Yes - both of those!

tilmanschweitzer:
I have the same issue and I can verify that the fix in the open pull request fixes my issue. Would be great to see that fix in an upcoming release.

yennor:
I've got the same problem. The patch #1513 fixed the problem when the user selects something and then reopens the dropdown. But if there was a value set using the v-model before the user the first time opens the drop down, it still doesn't scroll to the selected value.

astyltsvig:
I have still the same problem

The v-select wont scroll to the predefined selected value by first click

Here is a example in codepen.
https://codepen.io/Hoxi/pen/QWBXwVB?editors=1010

astyltsvig:
By one comment from stack-overflow, i've solved my issue.

'menu-props="auto" ' 

```
<v-combobox
  v-model="select"
  :items="items"
  menu-props="auto"
  label="Select"
  persistent-hint
  return-object
  single-line>
 </v-combobox> 
```

PR #1173, opened 2020-04-24
Fix Issue #884 update typeAheadPointer when dropdown open.

ocomsoft:
This fixes issue #884 by setting the typeAheadpointer to the index then the Drop down is open and scrolling into position.

### DIFF
--- a/src/mixins/typeAheadPointer.js
+++ b/src/mixins/typeAheadPointer.js
@@ -13,7 +13,24 @@ export default {
           break;
         }
       }
-    }
+    },
+    open(newValue) {
+      if (newValue === true) {
+        // Just Opened dropdown
+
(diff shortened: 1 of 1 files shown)


ocomsoft:
I believe this fixes issue #1169 also.

amulchinock:
> I believe this fixes issue #1169 also.

Not sure it does. Currently using version 3.11.2 - and there is no `.vs__dropdown-option--selected` class applied to any list options on initially opening. Will add this information to #1169 also.

amulchinock:
@sagalbot is there an ETA for merging this in? It’s been sat here for over a year now - and addresses multiple problems.

If you need volunteer help with maintenance, let me know. Would love to get some of these things unstuck. 😊

PR #1210, opened 2020-06-13
do not set $data._value in updateValue if reduce option is set, it is…

### LINKED ISSUE #1209: Changing v-model does not sync with vue-select internal value
**Describe the bug**

Here is a screen recording of the bug: https://www.loom.com/share/9830f13bce2540ae96d270d615083e58

What's happening is that when I click on the Add New Driver button, I call a function to make a new driver, and then also reset the `driverPhone` `v-model` prop to be null, so that should clear the v-select input field. (you can see this in the code below)

But this clearing only works the first two times. After that, the Add New Driver option appears to be stuck there and the `@input` event is not fired anymore until I select a different option before selecting the Add New Driver option again.

**To Reproduce**
```
<v-select
          v-model="driverPhone"
          @input="driverInputChanged"
          class="<omitted>"
          :options="driversSelectFmtd"
          label="nameWithPhone"
          :reduce="driver => driver.phoneNumber"
        >
          <template #no-options>
          <omitted>
          </template>
        </v-select>

...

export default Vue.Component("AddDelivery", {
..
data() {
    return {
      driverPhone: '',
    };
  },
methods: {
    driverInputChanged(val) {
      console.log(`\n\ndriver phone changed: ${val}`);

      // User clicked on adding a new driver
      if (val === ADD_DRIVER_OPTION) {
        // addNewDriverFunc();
        this.driverPhone = null; // this should reset the value of the v-select, but it isn't working when the same option is chosen repeatedly
      }
    },
...
}
```

**Expected behavior**
I expect that since I'm setting `this.driverPhone = null;` that I should be able to continually click on Add New Driver and it should always fire that event.

MacOS, Chrome Version 83.0.4103.97 (Official Build) (64-bit)
vue-select 3.10.3

schelmo:
… set by the watcher of value

if the parent denies to set an option, vue-select will still have the value stored (setted in updateValue()).
i also extended the v-model test in the Reduce spec

maybe fixes #1209 (could u test @himat ?)

---

Closes #1209

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -713,7 +713,7 @@
        * @param value
        */
       updateValue (value) {
-        if (this.isTrackingValues) {
+        if (typeof this.value === 'undefined') {
           // Vue select has to manage value
           this.$data._value = value;
         }
@@ -989,7 +989,6 @@
        */
       selectedValue () {
         let value = this.value;
-
         if (this.isTrackingValues) {
           // Vue select has to manage value internally
           value = this.$data._value;
--- a/tests/unit/Reduce.spec.js
+++ b/tests/unit/Reduce.spec.js
@@ -100,19 +100,36 @@ describe("When reduce prop is defined", () => {
     expect(Select.vm.selectedValue).toEqual([]);
   });
 
-  it("can use v-model syntax for a two way binding to a parent component", () => {
+  it("can use v-model syntax for a two way binding to a parent component", async () => {
     const Parent = mount({
       data: () => ({
         reduce: option => option.value,
-        value: "foo",
+        current: "foo",
         options: [
           { label: "This is Foo", value: "foo" },
           { label: "This is Bar", value: "bar" },
           { label: "This is Baz", value: "baz" }
         ]
       }),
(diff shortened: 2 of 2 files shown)


himat:
@schelmo how do I go about testing this? I made your changes to my locally installed `vue-select` package in my `node_modules/` of my project, but the changes don't seem to show up. Do you recommend using `yarn link` to try the local module changes or something else?

himat:
Here's what I tried, but I don't see the changes working when I import it in my project.
```
mkdir del/
cd del
yarn add vue-select
cd node_modules/vue-select
yarn link
<then I added a console.log in the src/components/Select.vue>
```
Then,
```
cd my-proj
yarn link vue-select
```
Then I ran my project, but I don't see the console.log being printed out.

schelmo:
the dist files are missing
use this in your projects package.json
"vue-select": "https://github.com/schelmo/vue-select.git#tmp-build",
then run npm install (or yarn install)

himat:
Ah great! It has fixed my problem in #1209 - thanks for the speedy help!!

Issue #1145, opened 2020-03-31
Option to avoid focus on clear button (tabindex) during keyboard navigation


krishnakumarb:
**Is your feature request related to a problem? Please describe.**
When I have _clearable_ set to true and navigating using tab button on the keyboard, focus moves from the dropdown to the clear button before moving to the next form element. 

**Describe the solution you'd like**
An option to remove tabindex on the clear button to that focus moves directly to the next input element.

ipetrov87:
Hi. I have the same issue, and it is critical for me. Is it there some information when do you think to implement and release fix of this issue?

wrurik:
Hello. Are there any plans for merging this feature in the near future?

mam805:
This is still a problem for me. Any updates or fixes in sight?

PR #1215, opened 2020-06-20
feat: add clearButtonTabIndex prop

markacianfrani:
This should solve https://github.com/sagalbot/vue-select/issues/1145#issuecomment-645836997 

I couldn't figure out the significance between the Vue update and this issue coming up, but I made this PR anyway because it seemed like something you might want to control. This is also keeping one of my projects locked on Vue 2.5, believe it or not. 

Alternatively, we could just have the tabindex on the clear button be -1 and not add any unnecessary props. 

And unfortunately, I just can't get my head around testing and Vue. Let me know if you would want a test for this and I'm sure I could figure something out.

### DIFF
--- a/docs/api/props.md
+++ b/docs/api/props.md
@@ -339,6 +339,17 @@ tabindex: {
 },
 ```
 
+## clearButtonTabIndex
+
+Set the tabindex for the clear button.
+
+```js
+clearButtonTabIndex: {
+	type: Number,
+	default: -1
+},
+```
+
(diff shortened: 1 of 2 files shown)


nieka:
Wil this pull request be merge in the near future? Or are there reasons to not merge it?

RobertDessin:
Would really appreaciate it, or at least set the clear-buttons tabindex to -1.

Issue #1044, opened 2020-01-21
How to set name for v-select element?


daphino:
how i can set name of this element when post data.

![Screenshot_11](https://user-images.githubusercontent.com/53916093/72795909-facd8c80-3c70-11ea-8cd6-6250bc18ab8a.png)

sagalbot:
Proxy the value to a hidden input:

```html
<v-select v-model="selected" />
<input type="hidden" :value="selected" />
```

Issue #676, opened 2018-10-09
How can I use this to just populate a simple input value?


bkuhl:
My objective is to use this component on a site that is only partially VueJS.  Ideally `<v-select>` in my case will populate a hidden in put field behind the scenes.

I see many references to using this with `v-model` and the docs even suggest you can use it without, but i'm not able to get this value to be part of the normal POST on my site as would a hidden input field.

Do I need to write my own component to make this populate a hidden input field or can it just be done as simply as using `<v-select>`?

gordonbisnor:
I was happy to see that Vue allows for this to work with the multiple select as well, eg:
```js
 <input type="text" :value="foo.map(m =>m.value)" />
```

sagalbot:
```html
<v-select v-model="selected" />
<input type="hidden" :value="JSON.stringify(selected)" />
```

You could use this approach to proxy the value of the `v-select` to a hidden input, and send it to the server.

sagalbot:
Thanks @geri777 for the support!

Issue #642, opened 2018-09-12
send value to hidden inputbox


pcm0nk:
Hi, im trying to send a selected option into an hidden input box but i get [object object] in return
i tried this solution : https://github.com/sagalbot/vue-select/issues/495 but it did not work

```
  <v-select v-tooltip.right="msg"  taggable max-height="100px" ref="jobbox" push-tags v-model="selected" dir="rtl" label="alljobstitle" name="selected"  :options="alljobsnew" >
</v-select> 
  <input type="hidden" name="selectedItem" :value="selected">

```
Laravel output
```
array:3 [▼
  "_token" => "FJ1m43y3Mm5CgwL2Q8meX1HTycMxTKCc9gl"
  "selectedItem" => "[object Object]"
  "ment" => "0"
]
```
i did try v-model still the same output, im stuck and out of time
thanks for any help!

sagalbot:
Hmm.. I'm taking a short in the dark here but maybe try converting the object into a string?

```
<v-select v-model="selected />
<input type="hidden" :value="JSON.stringify(selected)" />
```

pcm0nk:
Thanks dude u are a awsome

sagalbot:
Glad it worked!

prasetyo-jeanu:
not worked **selected** still same

sagalbot:
@prasetyo-jeanu please stop commenting on multiple issues.

PR #1246, opened 2020-08-01
feat: hidden input

boycce:
Setting `hiddenInputName: "name"` will render a hidden input field along with the dropdown:

```html
<input name="name" type="hidden" class="vs__hidden-input" value="['apple', 'orange']"  />
```

Demand: 

https://github.com/sagalbot/vue-select/issues/642
https://github.com/sagalbot/vue-select/issues/676
https://github.com/sagalbot/vue-select/issues/1044
...

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -77,6 +77,7 @@
       </ul>
       <ul v-else :id="`vs${uid}__listbox`" role="listbox" style="display: none; visibility: hidden;"></ul>
     </transition>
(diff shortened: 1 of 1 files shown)


src/components/Select.vue
```diff
@@ -77,6 +77,7 @@
       </ul>
       <ul v-else :id="`vs${uid}__listbox`" role="listbox" style="display: none; visibility: hidden;"></ul>
     </transition>
+    <input v-if="hiddenInputName" v-bind="scope.hiddenInput" type="hidden" class="vs__hidden-input">
```
Hagith, line 80: Isn't a `:name="hiddenInputName"` attribute missing here?

PR #1262, opened 2020-08-15
feat: Set search value to option selected if sync prop is given

### LINKED ISSUE #1188: Keep value in input on blur event, and additional options
First off all I would like to thank you for making this awesome plugin, and for providing consistent  support to its users :)

Yes

Hi Im using your search like this:

```
   <v-select
                    class="select-search"
                    :value="value"
                    :options="options"
                    @search="onSearch"
                    ref="searchSelect"
                    :placeholder="$t('label.select_action')"
                >
                    <template #no-options>Empty list</template>
                </v-select>


const options = ['first', 'second', 'third', 'first-1'];
```
What I would like it to work more like a autocomplete 

So If Im doing a search on input,( typing letters), on blur I would like my input field to keep input value inside v-select but dropdown can be closed. Like here:
http://jsfiddle.net/bebarr89/tZDAd/

And if I select value from dropdown then that value to be in input. Is that even possible ?

jijojames18:
Fixes #1188 
Added new prop `syncSearchValue`.
In single select, if the prop is set, then 

-  when the user clicks on an option from dropdown, the search value is set to the value of that option.
-  On mount, if a default value is passed in, then that value will be set as the search value.
- Clearing the select will clear the search as well.

This fix along with `clearOnSelect` prop can be used to implement the required feature mentioned in #1188

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -540,6 +540,18 @@
       },
 
       /**
+       * Sets the value of the currently selected option
+       * onto the input box
+       *
+       * Applicable only for single select
+       * @type {Boolean}
+       */
+      syncSearchValue: {
+        type: Boolean,
+        default: false
+      },
+
+      /**
        * When `appendToBody` is true, this function is responsible for
        * positioning the drop down list.
        *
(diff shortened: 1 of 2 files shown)


abrunovski:
Hi @jijojames18, is there was any reason to remove this prop?
It was a helpful feature.

Thanks

jijojames18:
Hi @abrunovski This feature is not yet merged and deployed.

PR #1289, opened 2020-09-24
Vue 3 Compatibility

ParaBolt:
+ Add emits property to list out all events that are being emitted.
+ Replace this.$on with a direct function call
+ Rename $scopedSlots to $slots
+ Rename value to modelValue
+ Change input event emit to update:modelValue event emit
+ Replace v-for refs with document.querySelector (Better Fix : Use a dynamic ref generation code in component)
+ Replace this.$options.propsData.hasOwnProperty('reduce') with typeof this.reduce === 'function' (Not sure about the correct approach)

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -100,6 +100,8 @@
 
     directives: {appendToBody},
 
+    emits : ['open', 'close', 'search:focus', 'search:blur', 'update:modelValue', 'option:created'],
+
     props: {
       /**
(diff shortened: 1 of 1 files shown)


alaamebrahim (approved):

ParaBolt:
Yo, welcome! Thanks again for the awesome plugin :)

VictorSabido:
Thanks @ParaBolt, Do we know when they will put him in master?

Issue #1307, opened 2020-11-03
Vue 3 compatibility


pud1m:
Hey, guys.

Vue 3 has been out for a while now and this plugin still isn't compatible. [This PR](https://github.com/sagalbot/vue-select/pull/1286) fixes compatibility but it still hasn't been merged into the main branch.

this.$on is breaking compatibility and makes the plugin unusable for people on Vue 3.

Is there a reason why it's still not merged yet?

wolfjagger:
Related: #1251.

Wish I had seen that PR before I did similar work. I have a fork with updates that works as far as I can tell: https://github.com/wolfjagger/vue-select. #1286/#1289 doesn't change webpack or tests as necessary. I'll make a PR from my fork, since I just got it working how I need.

sagalbot:
@pud1m the reason is time – I have none available at the moment. I freelance full time and have scheduled ~20 hours in December to do a round of updates for the component, Vue 3 compatibility is first on the list. If you'd like to see the update sooner, sponsorships really do help me dedicate time to open source that'd normally go to client projects.

bci24:
Any updates ? :)

PR #1286, opened 2020-09-18
fix(vue3): Vue 3 Compatibility

sagalbot:
- [x] drops vue 2.x peerdep
- [ ] implement global API updates ex) `Uncaught TypeError: this.$on is not a function`

Issue #1251, opened 2020-08-07
Vue 3 support


koljada:
**Is your feature request related to a problem? Please describe.**
It would be great to have a working version for Vue 3.

**Describe the solution you'd like**
New version with Vue 3 support.

**Describe alternatives you've considered**
-

**Additional context**
At the moment I can't even install the lib: 

> vue-select@3.10.7" has incorrect peer dependency "vue@2.x"

---

## Edited by @sagalbot: Vue 3 support is available on the `beta` release channel.

> Please see #1597 as the source of truth for the `v4` / Vue 3 compatible release. Thanks! - @sagalbot 

```
npm install vue-select@beta
```

```
yarn add vue-select@beta
```

wolfjagger:
I have a repo that depends on vue-select that I've been upgrading to vue 3. I forked vue-select (https://github.com/wolfjagger/vue-select) and upgraded as best I could.

Seems to install correctly, and unit tests now pass without too much modification. I don't know the library top-to-bottom, so I didn't comprehensively test manually. I didn't upgrade or remove more packages than necessary, so yarn audit's not great. I did need to install typescript; I think maybe @vue/test-utils needed it?

Unfortunately, there is a problem when using this as a dependency. When I import it in another package and try to mount it, it refuses and throws a bunch of warnings, starting with:

```
[Vue warn]: Invalid VNode type: Symbol() (symbol) 
  at <VueSelect ref="selectEl" class="select unfocused" modelValue=""  ... >
  ...
```

I don't know why this is; suspect a vue 3 bug. I see a couple of similar error messages on searching, and this seems very relevant: https://forum.vuejs.org/t/vue-3-invalid-vnode-symbol-fragment-when-using-an-external-component-with-a-slot/105448. May be a bug with slots, which the library relies on a lot.

Anyway, try it out if you want! And if you know or figure out the "Symbol" bug, please reply. :)

wolfjagger:
Took me way too long, but found the bug to be from bundling vue into the library. Vue was bundled in, so when the slot was rendered, it used a `Symbol(Fragment)`. Symbols are unique by default, so the vue that was bundled in and the vue that was used in a consuming library created two different `Symbol`s, and they were compared when trying to render. Since they did not show as equal, it refused to render the `Fragment` properly and emitted this error.

Fix is just adding `externals: { vue: 'vue' }` to webpack. Bonus is that the bundle size is smaller.

josh7weaver:
[...]

PR #1309, opened 2020-11-04
Vue 3 upgrade

wolfjagger:
Related issues & PRs I can find: #1251 #1286 #1289 #1307.

Upgrad vue to 3.
Upgrade vue-related packages. This uses several beta packages, so I don't know how we feel about that.
Event API breaking changes (emits, modelValue, ...).
Change 'propsData' to 'props'.
Change 'scopedSlots' to 'slots'.
Update dev.js with createApp breaking change.
Adjust webpack build for vue 3 (change vue alias to esm-bundler, add environment define variables).
Set vue as external library so it's not bundled in.
Add prepare script (can remove, just was helpful for testing).


wolfjagger:
There are two changes #1289 makes that I do not:

Select.vue:748 they change from `this.$refs['deselectButtons']` to `document.querySelectorAll('.vs__deselect')`.
Select.vue:985 they change from `this.$options.propsData.hasOwnProperty('reduce');` to `typeof this.reduce === 'function';`.

I assume both of these changes are for good reason, but I have not looked into it.

j2L4e:
> Select.vue:985 they change from this.$options.propsData.hasOwnProperty('reduce'); to typeof this.reduce === 'function';.

`this.$options.props.hasOwnProperty('reduce')` always throws, because `$options.props` doesn't have a prototype. And if it worked, it would always evaluate to true, because the reduce prop is defined on the very component.
`typeof this.reduce === 'function'` checks if there's a function passed in as a prop. Which seems odd, because it has a default of () => {}.

Issue #1266, opened 2020-08-17
How to apply css to dropdown list when the prop appendToBody is true


eddivalen:
How can I apply some css to change background of the dropdown list, when I'm using the prop appendToBody, the div is at the end of the body, there is some way to add a class or replace the ID?

I'm trying to applying some dark style...

![image](https://user-images.githubusercontent.com/8193094/90443073-a2952680-e0a9-11ea-9056-bb6adaba96e4.png)

```
<v-select class="select-x" taggable v-model="value" append-to-body :options="options"></v-select>

```

eddivalen:
Can you please give me an idea to solve this @sagalbot

sagalbot:
The `.vs__dropdown-menu` class is still applied, you can target that for your css 👍🏻

eddivalen:
@sagalbot is there a way to add a unique class or ID?

heivo:
I think the problem is that when setting a class (like it is described [here](https://vue-select.org/guide/css.html#overriding-default-styles), this class is not applied to `.vs__dropdown-menu` in the body. So there is no way to style this individually for one rendered component.

uturnr:
I also would like to be able to add a dynamic class to the dropdown menu. I have multiple dropdown menu styles set in my wrapper component, and I need `appendToBody`, but it breaks my custom styles.

SchmidtDawid:
Custom class applied for dropdown will do the job

EricRabil:
My workaround is as follows, and prevents me from having to inject global CSS.

```js
mounted() {
  this.$refs.select.$on("open", async () => {
    await this.$nextTick();
    await this.$nextTick();

    this.$refs.select?._vnode.children.forEach(({ elm }) => {
      if (elm instanceof HTMLElement && elm.tagName.toLowerCase() === "ul") {
        elm.classList.add(this.remoteClassName);
      }
    });
  });
}
```

It'd be nice to have prop for a custom class-name though, so that we don't have to do hacks like this. The changes I want to do in this scenario are not changes I want to apply to all dropdowns.

alantea:
One more workaround is to use [calculatePosition with Propper.js](https://vue-select.org/guide/positioning.html#popper-js-integration), the dropdown menu element can be edited when adjusting the position.

```html
<v-select :options="countries" append-to-body :calculate-position="withPopper" />
```

```javascript
methods: {
  withPopper (dropdownList, component, {width}) {
    dropdownList.classList.add("style-chooser");
[...]

PR #1341, opened 2020-12-23
feat: events for append-to-body

sagalbot:
Provides an API for #1266

- adds two events: `dropdown:appending` & `dropdown:removing`
- both events receive an object with the dropdown menu `{ el: HTMLUListElement }`

This provides a hook to further adjust the element without losing the default position calculation.

```vue
<v-select @dropdown:appending="({ el }) => el.classList.add('my-custom-select')" />
```

### DIFF
--- a/src/directives/appendToBody.js
+++ b/src/directives/appendToBody.js
@@ -1,4 +1,12 @@
 export default {
+    /**
+     * called when the bound element has been inserted into its parent node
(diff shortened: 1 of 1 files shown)


ibam-voxel:
@sagalbot hello you must fix this.

because branch has conflict!

adamreisnz:
I don't think this ever ended up being implemented?

PR #1362, opened 2021-01-28
Add uniqueKey prop

### LINKED ISSUE #1360: UID's break Jest snapshots
**Version** 3.11.2

When updating to the latest version of Vue-Select, it included a change to the DOM output of unique ID's. These numbers are semi-random when ran in tests. The UID is incremented for every instance of the component in a test, and for every test that runs. Jest intentionally runs tests in different orders each time to ensure your tests are properly set up and not reliant on the outcomes of previous tests. Depending on the order of the test, the ID will be different from the last time it ran.

Jest offers a "Snapshot" feature that is useful to ensure no changes occur in the DOM. It stores a copy of the expected DOM and displays a diff of the expected and received snapshots:

```diff
  <td class="vgt-left-align">
    <div class="pmr-form-group mb-0">
      <div dir="auto" class="v-select vs--single vs--searchable vs--disabled">
-       <div id="vs250__combobox" role="combobox" aria-expanded="false" aria-owns="vs250__listbox" aria-label="Search for option" class="vs__dropdown-toggle">
+       <div id="vs258__combobox" role="combobox" aria-expanded="false" aria-owns="vs258__listbox" aria-label="Search for option" class="vs__dropdown-toggle">
          <div class="vs__selected-options">
            <span class="vs__selected">
              admin
              <!---->
            </span>
-           <input disabled="disabled" aria-autocomplete="list" aria-labelledby="vs250__combobox" aria-controls="vs250__listbox" type="search" autocomplete="off" class="vs__search">
+           <input disabled="disabled" aria-autocomplete="list" aria-labelledby="vs258__combobox" aria-controls="vs258__listbox" type="search" autocomplete="off" class="vs__search">
          </div>
          <div class="vs__actions">
            <button disabled="disabled" type="button" title="Clear Selected" aria-label="Clear Selected" class="vs__clear" style="display: none;">
              <svg xmlns="http://www.w3.org/2000/svg" width="10" height="10">
                <path d="M6.895455 5l2.842897-2.842898c.348864-.348863.348864-.914488 0-1.263636L9.106534.261648c-.348864-.348864-.914489-.348864-1.263636 0L5 3.104545 2.157102.261648c-.348863-.348864-.914488-.348864-1.263636 0L.261648.893466c-.348864.348864-.348864.914489 0 1.263636L3.104545 5 .261648 7.842898c-.348864.348863-.348864.914488 0 1.263636l.631818.631818c.348864.348864.914773.348864 1.263636 0L5 6.895455l2.842898 2.842897c.348863.348864.914772.348864 1.263636 0l.631818-.631818c.348864-.348864.348864-.914489 0-1.263636L6.895455 5z" />
@@ -189,11 +189,11 @@
              <path d="M9.211364 7.59931l4.48338-4.867229c.407008-.441854.407008-1.158247 0-1.60046l-.73712-.80023c-.407008-.441854-1.066904-.441854-1.474243 0L7 5.198617 2.51662.33139c-.407008-.441853-1.066904-.441853-1.474243 0l-.737121.80023c-.407008.441854-.407008 1.158248 0 1.600461l4.48338 4.867228L7 10l2.211364-2.40069z" />
            </svg>
            <div class="vs__spinner" style="display: none;">Loading...</div>
          </div>
        </div>
-       <ul id="vs250__listbox" role="listbox" style="display: none; visibility: hidden;"></ul>
+       <ul id="vs258__listbox" role="listbox" style="display: none; visibility: hidden;"></ul>
      </div>
    </div>
  </td>
```

Vue-Select already has an option to pass in an `inputId`, why not just use that in these places? Or have an optional `key` prop, and use that if provided, and if not, fall back on the existing UID approach.

This wouldn't be that difficult to implement, just a `computed` property called `uid` instead of a [`data`](https://github.com/sagalbot/vue-select/blob/master/src/components/Select.vue#L573). Something like this:

```js
props: {
  uniqueKey: {
    type: [String, Number]
  }
},
computed: {
  uid: function () {
    return this.uniqueKey || this.inputId || uniqueId();
  }
}
```

TheJaredWilcurt:
Fixes #1360

### DIFF
--- a/docs/api/props.md
+++ b/docs/api/props.md
@@ -544,6 +544,17 @@ transition: {
 ```
 
 
+## uniqueKey
+
+Optional unique key for use on DOM attributes.
+
+```js
+uniqueKey: {
+    type: [String, Number]
+},
+```
+
+
 ## value
 
 Contains the currently selected value. Very similar to a
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -470,6 +470,15 @@
       },
 
       /**
+       * Optional unique key for use on DOM attributes.
+       * @type {String}
+       * @default {null}
+       */
+      uniqueKey: {
+        type: [String, Number]
+      },
+
+      /**
        * Sets RTL support. Accepts 'ltr', 'rtl', 'auto'.
        * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir
        * @type {String}
@@ -570,7 +579,6 @@
 
     data() {
       return {
-        uid: uniqueId(),
         search: '',
         open: false,
         isComposing: false,
@@ -978,6 +986,13 @@
     },
 
     computed: {
+      /**
+       * Produce a unique ID to use in the DOM.
+       * @return {string}
+       */
+      uid () {
+        return String(this.uniqueKey || uniqueId());
+      },
       /**
        * Determine if the component needs to
        * track the state of values internally.


TheJaredWilcurt:
I'm hoping this (or something equivalent) can be merged/released, it is blocking us from upgrading this dep at work.

TheJaredWilcurt:
@sagalbot I know you're working hard on Vue 3 stuff, but would it be possible to squeeze this in to a Vue 2 release?

PR #1369, opened 2021-02-10
When selecting an option with Enter key and typeAhead it should be deselected if it is selected already

peda:


peda:
this is for v2

peda:
please ignore

peda:
please ignore

PR #1487, opened 2021-09-01
add appendTo option and mark appendToBody option as deprecated

rx-837:


rx-837:
@sagalbot what about pull request? As can i see still it not merged(

daniel-anh-nguyen:
@rx-837 @sagalbot I think this change to append-to is really useful. Would this be merged soon?

PR #1530, opened 2021-10-22
feat: add dropdownOptionWrap prop with 3 strategies

### LINKED ISSUE #1455: Line breaks in options slot
If having long option labels the dropdown starts to look bad because of the overflow. See the following picture:

image

It would look much better to have line breaks in the options slot by default and thus not having the horizontal overflow and scrolling. I believe that this is quite a common problem.

My current solution through the option slot is:
```vue
<template #option="option">
  <div style="white-space: normal; vertical-align: top; display: inline-block; line-break: loose; overflow-wrap: anywhere;">
    <p style="margin: 0;">
      {{ option.label }}
    </p>
  </div>
</template>
```

We are breaking lines with the 'loose' rule and if any word is longer than the whole inline-block, then we break it with the overflow-wrap: anywhere rule.

image

sagalbot:
Closes #1455

### DIFF
--- a/src/scss/modules/_search-input.scss
+++ b/src/scss/modules/_search-input.scss
@@ -48,6 +48,7 @@ $font-size: 1em;
 // Empty
 .vs--not-empty .vs__search {
     height: 0;
+    margin: 0;
 }
 .vs--not-empty .vs__search:focus {
     height: auto;


sebix:
The examples in https://deploy-preview-1530--vueselect.netlify.app/guide/css.html#option-wrapping-strategy ([via](https://github.com/sagalbot/vue-select/issues/1455#issuecomment-949906284)) indeed look very good. Thank you for your work on this!

PR #1534, opened 2021-10-30
fix(vue3): migrate refs array

gnuletik:
see https://v3.vuejs.org/guide/migration/array-refs.html#migration-strategy

Fixes `Invalid attempt to spread non-iterable instance` error when : 
- instantiating a vue-select component with `multiple`
- selecting an option
- opening dropdown

Related to https://github.com/sagalbot/vue-select/issues/1251


gnuletik:
Thanks for the release!

Issue #1532, opened 2021-10-26
[v4 beta] Invalid attempt to spread non-iterable instance when using multiple mode


Finrod927:
**Describe the bug**
When using the multiple mode with the new v4-beta.1, the component seems to work, the `v-model` is correctly updated but I have a warning and an error in the console when selecting a second value or removing a selected value:

> [Vue warn]: Unhandled error during execution of native event handler

> Uncaught TypeError: Invalid attempt to spread non-iterable instance
    at e.exports (vue-select.js?4a7a:1)
    at e.exports (vue-select.js?4a7a:1)
    at Proxy.toggleDropdown (vue-select.js?4a7a:1)
    at Object.onMousedown.t.<computed>.t.<computed> (vue-select.js?4a7a:1)

**To Reproduce**

```
<template>
  <VSelect id="multiple" v-model="multiple" :options="['Option 1', 'Option 3', 'Option 3']" multiple />
</template>

<script lang="ts">
import { defineComponent } from 'vue'
import VSelect from 'vue-select'

export default defineComponent({
  name: 'TestingVS4',

  components: {
    VSelect
  },

  data() {
    return { multiple: [] }
  }
})
</script>
```

1. select one option: no error
2. select a second option: error in console
3. remove a selected option: error in console

**Versions**

- vue: 3.2.20
- vue-select: 4.0.0-beta.1

roznik-navisys:
Before this gets fixed, you can filter out the error using Vue's built-in error handler.

```ts
app.config.errorHandler = (err, vm, info) => {
    // I know I could do a better job with the filtering
    if (err instanceof TypeError && err.stack?.includes('vue-select')) {
        console.log(
            'v-select beta error, remember to update the npm module later',
        );
    } else throw err;
};
```

dieg0v:
For now, I am doing the same capture that @roznik-navisys  does but inside the component:

```js
export default {
  errorCaptured(err, vm, info) {
    return false;
  }
...
}
```

Ali-Shaikh:
Came here for the exact error

bci24:
The same error..... any news about this fixed ? It's annoying when selecting items and get error by error in console :)

```
Uncaught TypeError: Invalid attempt to spread non-iterable instance
```

sagalbot:
Fixed in `4.0.0-beta.2` via #1534

PR #1543, opened 2021-11-22
Fix toggleDropdown bug

serkanerip:
Hi, when using multiple mode with beta version there is a bug when try to select after first select.

image

This bug comes from toggleDropdown function in select.vue component.

### Here my config for dev.vue to reproduce bug:
Screen Shot 2021-11-22 at 12 40 34

Type of this.$refs['deselectButtons'] is object after first selection in multiple mode and tries to spread on object.

```
const ignoredButtons = [
        ...(this.$refs['deselectButtons'] || []),
        ...([this.$refs['clearButton']] || []),
]
```

Actually, on the bottom line for clearButton ref its wrapped in an array. I applied same for deselectButtons.

This commit fixes the bug.

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -1101,7 +1101,7 @@ export default {
       //  don't react to click on deselect/clear buttons,
       //  they dropdown state will be set in their click handlers
       const ignoredButtons = [
-        ...(this.$refs['deselectButtons'] || []),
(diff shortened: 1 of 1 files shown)


serkanerip:
This pr fixes #1532.

aedart:
Is there any chance that this can be released soon? Could really use this patch / fix.

PR #1594, opened 2022-02-18
refactor: move from webpack to Vite

sagalbot:


src/index.js
```diff
@@ -1,5 +1,3 @@
 import VueSelect from './components/Select.vue'
-import mixins from './mixins/index'
 
 export default VueSelect
-export { VueSelect, mixins }
```
sagalbot, line 5: Breaking change

PR #1650, opened 2022-07-17
feat: add input-background css variable

KareemDa:
The goal for this PR is to be able to change the input background color.

Material Design has an option: `filled` which changes background color from `white` to `light` one, 

currently, the background is hardcoded. 
This PR adds new variable: `--vs-input-background` to override default background color

Example of `filled` applied by customizing `--vs-input-background`:

image

### DIFF
--- a/src/css/global/variables.css
+++ b/src/css/global/variables.css
@@ -6,6 +6,7 @@
 
     /* Search Input */
     --vs-search-input-color: inherit;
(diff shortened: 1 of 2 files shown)


KareemDa:
Thank you so much!

JoA-MoS:
@sagalbot - could this functionality be merged into the beta channel for vue3 support?

PR #1655, opened 2022-07-21
Provide a slot for clearButton

vstyler96:
In this PR we are just including a slot for the clearButton section in case our needs don't fit well with the current conditional schema for showing up the clear button.


moisish:
Any chance that this will be released soon?

jeanlrnt (approved):

PR #1688, opened 2022-09-16
feat: add aria label props

gavmck:
Added aria labels as props, both so we can support them being more customised to the context and for i18n support.


shaunkrd (approved):
Love it, thanks. We need these additions!

jeanlrnt (approved):
This is a great addition for accessibility. Good job!

PR #1726, opened 2022-12-14
Add aria-multiselectable

Pytal:
Enhance component accessibiltiy as described in the MDN docs:

- [aria-multiselectable Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable)
- [aria-selected Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected)

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -91,6 +91,7 @@
         v-append-to-body
         class="vs__dropdown-menu"
         role="listbox"
+        :aria-multiselectable="multiple"
         tabindex="-1"
         @mousedown.prevent="onMousedown"
         @mouseup="onMouseUp"
@@ -109,7 +110,7 @@
             'vs__dropdown-option--highlight': index === typeAheadPointer,
             'vs__dropdown-option--disabled': !selectable(option),
           }"
-          :aria-selected="index === typeAheadPointer ? true : null"
+          :aria-selected="optionAriaSelected(option)"
           @mouseover="selectable(option) ? (typeAheadPointer = index) : null"
           @click.prevent.stop="selectable(option) ? select(option) : null"
         >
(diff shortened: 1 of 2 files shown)


Pytal:
Tests added!

src/components/Select.vue
```diff
@@ -1212,6 +1213,24 @@ export default {
...
+     * @return {null|boolean}
+     */
+    optionAriaSelected(option, index) {
+      if (!this.selectable(option)) {
+        return null
+      }
+      if (this.multiple) {
+        return this.isOptionSelected(option)
+      }
+      return index === this.typeAheadPointer ? true : null
```
sagalbot, line 1231: I think my original functionality here was actually incorrect. I was using `aria-selected` to denote that the option is currently highlighted, but after rereading the docs, it should be used to denote the current selection. I think it should probably be `this.isOptionSelected` in all cases. 

Let me know if you think I'm incorrect there.

sagalbot (changes requested):

src/components/Select.vue
Pytal: Agreed and updated!

Also converted the boolean to string to match the docs https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#values with updated tests!

Pytal: Look good now @sagalbot?

AndyScherzinger:
Hi @sagalbot ,
first of all let me say thank you for all the work you are doing in offering this component, this is highly appreciated and we love your work. Not wanting to put any pressure on you just wanting to ask to have a clear view on a timeline-dimension, do you have a gut-feeling if or when you might be able to take another look at this PR and @Pytal's changes based on your latest review? Same goes for #1742. And again no pressure, it is a privilege to be able to use this component you are building and maintaining ❤️

PR #1727, opened 2022-12-14
fix: Keep focus on input after select to improve accessibility

Pytal:
Keep focus on the input after option selection and when hitting the Escape key to comply with W3C accessibility guidelines as demonstrated and described in:

- [Single-select example](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html)
- [Multi-select example](https://www.telerik.com/kendo-angular-ui/components/dropdowns/multiselect/accessibility/)

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -960,7 +960,7 @@ export default {
     },
 
     search(search) {
-      if (!this.open && search) {
+      if (search.length) {
         this.open = true
       }
     },
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -958,6 +958,12 @@ export default {
     open(isOpen) {
       this.$emit(isOpen ? 'open' : 'close')
     },
+
+    search(search) {
+      if (!this.open && search) {
+        this.open = true
+      }
+    },
   },
 
   created() {
@@ -1350,22 +1356,6 @@ export default {
 
       if (typeof handlers[e.keyCode] === 'function') {
         return handlers[e.keyCode](e)
-      }
-
-      const isModifierKeyActive = (e) => {
-        // Allow `shiftKey` for capitalized characters
-        const modifierProperties = ['ctrlKey', 'altKey', 'metaKey']
-        return modifierProperties.some((property) => e[property])
-      }
-
-      const alphanumericCharRegex = /^[a-z0-9]{1}$/i
-
-      if (
-        !this.open &&
-        !isModifierKeyActive(e) &&
-        alphanumericCharRegex.test(e.key)
-      ) {
-        this.open = true
       }
     },
   },
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -1239,7 +1239,7 @@ export default {
      */
     onEscape() {
       if (!this.search.length) {
-        this.searchEl.blur()
+        this.open = false
       } else {
         this.search = ''
       }
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -571,7 +571,12 @@ export default {
      */
     selectOnKeyCodes: {
       type: Array,
-      default: () => [13],
+      default: () => [
+        // enter
+        13,
+        // space
+        32,
+      ],
     },
 
     /**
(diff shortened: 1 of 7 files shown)


Pytal:
> Thanks for this! This behaviour is definitely how the component should work. This needs some refining though before I can merge. After a selection has been made, the input still has focus, but it's unresponsive. There's no way to bring the dropdown back up after a selection was made since the open state is tied to input focus.
> 
> I think what likely needs to happen, is the outer div of the template gets `tabindex="0"`, and when it is focused and a key is pressed, the dropdown opens and if `filterable=true`, focus is moved to the input box. I am fixing this in the next major release, but if you'd like to take a try at this, please do.

No problem and thanks for taking a look! I've tried reproducing this locally and on https://deploy-preview-1727--vueselect.netlify.app/sandbox.html but both seem to work for me as after selecting an option I can hit Enter to reopen the dropdown and make another selection. Which steps can I take to reproduce the unresponsiveness?

Pytal:
Made the changes and added some tests, what do you think @sagalbot?

src/components/Select.vue
```diff
@@ -1339,6 +1351,22 @@ export default {
...
+
+      const alphanumericCharRegex = /^[a-z0-9]{1}$/i
+
+      if (
+        !this.open &&
+        !isModifierKeyActive(e) &&
+        alphanumericCharRegex.test(e.key)
+      ) {
+        this.open = true
+      }
```
sagalbot, line 1369: I wonder if we could achieve the same effect via the `search` watcher? That would basically defer the complexity of the regex to the HTML search input itself. If the search value changes, and has a `length > 0`, then we know there's a value present in the search input. Given the tests you added, should be simple to check if that works.

sagalbot (changes requested):

src/components/Select.vue
Pytal: Added a commit with the `search` watcher approach but now this test is failing:
```
 FAIL  tests/unit/TypeAhead.spec.js
  ● Moving the Typeahead Pointer › should set the pointer to zero when the filteredOptions watcher is called

    expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: -1

      15 | 
      16 |     await Select.vm.$nextTick()
    > 17 |     expect(Select.vm.typeAheadPointer).toEqual(0)
         |                                        ^
      18 |   })
      19 | 
      20 |   it('should move the pointer visually up the list on up arrow keyUp', () => {
```

Any pointers?

sagalbot: That one took a while to figure out. A bit of a disadvantage of the watcher approach, it can be difficult to trace, especially across mixins. The conflict is in `mixins/typeAheadPointer.js`:

```js
  watch: {
    ///

    open(open) {
      if (open) {
        this.typeAheadToLastSelected()
      }
    },
    
	///
  },
```

`typeAheadToLastSelected()` was forcing the pointer to `-1`. Fixed by conditionally setting the pointer value only if there was a match found.

src/components/Select.vue
```diff
@@ -571,7 +571,12 @@ export default {
      */
     selectOnKeyCodes: {
       type: Array,
-      default: () => [13],
+      default: () => [
+        // enter
+        13,
+        // space
+        32,
```
hindarwan, line 578: I think space must not default. Consider if we want to search an option that contain space.

sagalbot: Thank you for catching this, 100% correct. I'll address this asap.

sagalbot: Fixed in 3.20.2, thanks @hindarwan for flagging!

YuryShautsouGY:
Hi all! You are doing an excellent work.

But we have one problem with this new behavior after the 3.20.1 updates.

On the input event, we call the API and open the bootstrap-vue b-modal dialog before the endpoint call completes. **After closing the modal dialog, the dropdown list reopens**, and this new behavior is something we don't expect at all. **How can we leave the dropdown menu closed** after closing the modal dialog. 

Maybe new props need to be made to prevent this behavior?!

Thanks!

PR #1745, opened 2023-01-19
fix(appendToBody) use vue built-in Teleport

### LINKED ISSUE #1294: Duplication of dropdown `appendToBody` and Vue-Router
**Describe the bug**
The dropdown is not deleted after each switching page with Vue-Router.
So multiple dropdown are maintain on the page.

**To Reproduce**
1. I apply `append-to-body` and `:calculate-position="withPopper"` like in the guide : [Dropdown position](https://vue-select.org/guide/positioning.html#calculated)
2. When I use the select, a listbox is created
3. I change the page and return back
4. A new listbox is created

**Expected behavior**
Create and use one dropdown

**Screenshots**
image

 - OS: Windows
 - Browser Chrome
 - Version 86.0.4240.75

s-montigny-desautels:
The teleport component will move the menu to the body and do a proper cleanup.
We could also expose the "to" props to be configurable by the end user.

It was not a problem with vue-router per-say, but more a problem when the component was unmounted.

fix #1294

### DIFF
--- a/dev/Dev.vue
+++ b/dev/Dev.vue
@@ -1,6 +1,7 @@
 <template>
   <div id="app">
-    <v-select v-model="selected" v-bind="config" />
+    <button v-if="config.appendToBody" @click="hide">Hide</button>
+    <v-select v-if="!hidden" v-model="selected" v-bind="config" />
   </div>
 </template>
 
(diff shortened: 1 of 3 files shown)


greendrake:
I suggest this alternative approach: https://github.com/sagalbot/vue-select/pull/1775

PR #1823, opened 2024-03-05
Feature, Limit a new property added 

gautam-patadiya:
To restrict display of selected options into selected container. 
A new property called 'limit' introduced


morganp-dev:
@sagalbot is there a plan to merge this PR or implement a similar feature for limited the number of displayed selections?

